PR #29, opened 2014-07-14
Perturb mineral

sannecottaar:
I wrote a little perturb mineral thing and modified step_3 from the tutorial to example_uncertainties.py to use it. What do you guys think? And what do you think about removing 'err_' from all the uncertainty dictionaries?


bobmyhill:
Can this PR be closed now, Sanne? Or do you still want to use some of it?

PR #95, opened 2014-12-28
one way to add molar_mass for SolidSolution

sannecottaar:
The new solid solution didn't have a molar_mass defined. Here is a proposal to have molar_mass as an attribute of Mineral and not as a function, which I think might be the more correct way, but maybe not the most user-friendly. Other ways without removing Mineral.molar_mass() might be:
- define a params dictionary for SolidSolution
- put an if statement for SolidSolution in Mineral.molar_mass()
neither of which I think is pretty. Unfortunately, what is not very elegant in my solution is that Mineral.molar_mass is now defined in the set_method function (cannot be done in __init__ in case someone defines an empty Mineral())

As a side, I commented out a test in test_spin.py that used a pt_dependent mineral, as these now fail and we have already excluded other examples because of this.


sannecottaar:
I haven't been writing tests so far, is this the idea?
(squashing the commits is giving me headaches)

bobmyhill:
Yeah, this is the idea. I might have the average of the molar masses as the comparator, rather than hardwiring a value.

Is there still an indent issue for the new test?

sannecottaar:
Should be better now.

bobmyhill:
Excellent :) ... I think indents may need to be fixed in solidsolution and maybe other places too... I think this may have been my fault originally... my emacs is set up for a three space indent, rather than four...

tjhei:
Sanne, can you please clean up this PR by rebasing to master and squashing the commits?

burnman/solidsolution.py
```diff
@@ -64,6 +64,13 @@ def set_composition(self, molar_fraction ):
         assert(sum(molar_fraction) < 1.0001)
         self.molar_fraction = molar_fraction 
 
+    def molar_mass(self):
+        """
+        Returns molar mass of the mineral [kg/mol]
+        """
+	molar_mass = sum([ self.base_material[i][0].molar_mass() * self.molar_fraction[i] for i in range(self.n_endmembers) ])
```
tjhei: indentation is messed up here

PR #99, opened 2014-12-31
MT split into isothermal and thermal

bobmyhill:
Opened for comments


ian-r-rose:
Is this safe to merge now?  Are you happy with the names 'isomt' and 'mtait'?

bobmyhill:
It should be safe, *but* I don't like those names. I didn't want to break things before at least one other person was happy.

What do we think of 'mtait' for the Modified Tait and 'hp_tmt' for the Holland and Powell thermal extension?

tjhei:
So "hp_tmt" refers to ```hp.HPMT``` and "mtait" to ```mt.MT```? I would prefer if the class and string name are the same.

bobmyhill:
Agreed... is this better?

tjhei:
Yes. I don't know if the chemistry makes sense ( @ian-r-rose ?).

It would be awesome if we had some unittests at some point that exercise the methods in the MT classes.

tjhei:
(tests are working for me)

ian-r-rose:
Looks fine to me.  'tmt' stands for thermal modified tait?

bobmyhill:
Yes :)

tjhei:
Can you add that to the class documentation please?

bobmyhill:
Ok, this should be good to go now...

burnman/mineral.py
```diff
@@ -64,6 +65,15 @@ class which derives from the equation_of_state base class.
             raise Exception("Please derive your method from object (see python old style classes)")
 
         if self.method is not None and type(new_method) is not type(self.method):
+            # Reset everything but params
+            for key, elem in self.__dict__.items():
+                if key != 'params':
+                    elem = float('nan')
+
```
ian-r-rose: What is the intention here?  I am confused...

ian-r-rose: This seems pretty dangerous.  Can we guarantee that there are no other members of the mineral that will be broken by this?  I'd rather not have it, I think.

bobmyhill: Well, the problem is that if we declare a new set_method, all of the old values of density, Vp, Vs are maintained. Some people might assume that properties are automatically recomputed. I'd rather not give anyone the change to misunderstand.

Even worse, if we try to set_state after set_method with the old P and T, your very useful catch to not recompute values if P and T is the same will end up confusing/misleading a lot of people.

If you don't like this method, I can suggest two alternatives:
1) Remove the P and T params only
2) Remove nothing, but output an additional warning, and store the old method as a temporary attribute that is deleted on the first new call of set_state.

ian-r-rose: I still think that the way this is written is way too dangerous.  For instance, will this set self.method to NaN?  What if somebody subclasses mineral and adds a few new members that are not P,T dependent?

As an alternative, take a look at what Timo has done in the performance improvements branch (i.e., manually setting the derived properties to None).  This is more explicit and less likely to break things in unexpected ways.

bobmyhill: Ok, let me try to get something together tomorrow that will be more explicit. I agree with your points, obviously what I've done here isn't the best solution (although set method itself wouldn't be a problem, as it is reset in the following line).

tjhei:
My assumption is (I guess it is not documented though) that you need to call set_state after changing the method. Should that not take care of this problem automatically?

ian-r-rose:
I think you are right in your assumption, Timo.  I guess the point is to protect the user from calling set_method() and then not calling set_state() thereafter.  I will add a note to this effect in the doc_updates branch.

I am coming around on the setup that you have done in the performance optimizations branch, so eventually we could do the same thing here.

bobmyhill:
Ok, this is certainly better. It's still not perfect (it doesn't reset parameters which are not contained in the new base class), but I think this is an issue for a different branch, after we make a decision on restructuring.

tjhei:
I still don't like that you are calling ```set_state``` inside ```set_method```. It is much nicer to have the program blow up in your face if you don't call ```set_state``` yourself. So maybe just reset P and T inside ```set_state```? In the long run it would be a good idea to have an internal ```reset_state``` method in ```Material``` that can be overridden.

bobmyhill:
Better?

ian-r-rose:
Of course, resetting P and T to NaN does not change the issue of other stale derived parameters.  I guess I'm not completely sold on it being a big issue, though.

tjhei:
we can fix that in a separate PR with the performance improvements, Ian.

tjhei:
Why are we changing ```self.params['equation_of_state']``` in ```set_method```?

burnman/mineral.py
```diff
@@ -64,7 +65,17 @@ class which derives from the equation_of_state base class.
             raise Exception("Please derive your method from object (see python old style classes)")
 
         if self.method is not None and type(new_method) is not type(self.method):
-            warnings.warn('Overriding database equation of state. From ' + self.method.__class__.__name__ + ' to ' + new_method.__class__.__name__, stacklevel=2)
+
+            # Warn user that they are changing the EoS
+            warnings.warn('Overriding database equation of state from ' + self.method.__class__.__name__ + ' to ' + new_method.__class__.__name__ + '. This does not overwrite any mineral attributes other than temperature and pressure. Non-implemented attributes of the old method will remain unchanged unless they are explicitly changed by the user.', stacklevel=2)
```
tjhei: I don't quite understand this new warning. What are you trying to say?

burnman/mineral.py
```diff
@@ -64,7 +65,17 @@ class which derives from the equation_of_state base class.
...
+
+            # Set new EoS
+            self.params['equation_of_state']=equation_of_state
```
tjhei: why do we need this part?

burnman/mineral.py
bobmyhill: If we change method from HP_TMT to MT, then it no longer makes any sense to have gibbs, Cp etc etc etc, but all of these attributes will be left unchanged from the previous set_state. This is a warning to the user so that they won't try to use any of thosee values after their next set_state. Eventually, we need to do as Ian suggests, but this warning is enough (IMO) for the current pull request.

burnman/mineral.py
bobmyhill: See the changes in misc/ref/example_compare_all_methods.py.out

tjhei: I don't quite get it, you want this so that the error message reflects the old method that was used? I don't think that makes sense. We want to warn if the user sets a method that is not the one for which the parameters got derived for (and this is tracked in params[eos]). So I vote against changing this.

bobmyhill: Ok, I see your point, but I wanted to cover the case covered by compare_all_methods. Can we cover both in the warning? (change from MGD3 -> MGD2, originally SLB2), or something similar?

I agree that the param shouldn't be changed, that was not my brightest idea.

tjhei: let's say I have a mineral with "SLB2" parameters and I now do
```
mat.set_method("mgd3")
...
mat.set_method("mgd2")
...
mat.set_method("slb2")
```

Will currently produce:
```
Warning: change from SLB2 to MGD3.
Warning: change from SLB2 to MGD2.
(no warning)
```

With your change we will get:
```
Warning: change from SLB2 to MGD3.
Warning: change from MGD3 to MGD2.
Warning: change from MGD2 to SLB2.
```

I think the last ```set_method``` is the "natural" method and should not produce a warning. Maybe we should rephrase it to say "Warning, you are changing the method to __ even though the material is defined as to be used with ___"?

ian-r-rose: I agree with Timo.  As we learned a couple of weeks ago, set_method() is a little bit fragile and needs to be treated with care.  We can try to cover Bob's case a bit more completely in the optimize branch, but this is a bit outside of the scope of the current PR.  Clarifying the warning seems fine, though.

bobmyhill:
Ok, I think this will make us all happy. I'm still slightly uncomfortable with leaving the set_method back to original unwarned without resetting all of the attributes, but we can fix this later. It's unlikely it would ever arise in practise anyway.

burnman/mineral.py
```diff
@@ -64,7 +65,17 @@ class which derives from the equation_of_state base class.
             raise Exception("Please derive your method from object (see python old style classes)")
 
         if self.method is not None and type(new_method) is not type(self.method):
-            warnings.warn('Overriding database equation of state. From ' + self.method.__class__.__name__ + ' to ' + new_method.__class__.__name__, stacklevel=2)
+
+            # Warn user that they are changing the EoS
+            warnings.warn('Warning, you are changing the method to ' + new_method.__class__.__name__ + ' even though the material is designed to be used with the method ' + self.method.__class__.__name__ + '.  This does not overwrite any mineral attributes other than temperature and pressure, which are set to nan. Stale attributes will be preserved unless they are refreshed (for example, by set_state).', stacklevel=2)
+
+            # Set new EoS
+            self.params['equation_of_state']=equation_of_state
```
ian-r-rose: I'm happy with the warning, though did we agree to remove this change to params?

bobmyhill: Crap, yes. Changed now.

ian-r-rose:
Okay, I'm happy with merging.  Any chance you could squash those last four commits before I do?

burnman/eos/hp.py
```diff
@@ -0,0 +1,330 @@
...
+        C_v = self.heat_capacity_v(pressure, temperature, volume, params)
+        K_S = K_T*C_p/C_v
+        return K_S
+
+    def gibbs_free_energy(self,pressure,temperature, volume, params):
+        """
+        Returns the gibbs free energy [J/mol] as a function of pressure [Pa]
+        and temperature [K].
+        """
+       # Calculate temperature and pressure integrals
```
tjhei: indentation is messed up

burnman/eos/modified_tait.py
```diff
@@ -3,312 +3,133 @@
...
 class MT(eos.EquationOfState):
     """
     Base class for a generic modified Tait equation of state.  
     References for this can be found in Huang and Chow (1974) 
     and Holland and Powell (2011; followed here).
+
+    An instance "m" of a Mineral can be assigned this 
+    equation of state with the command m.set_method('mt')
+    (or by initialising the class with the param 
+    equation_of_state = 'mt'
```
tjhei: closing parenthesis missing

tjhei:
Otherwise good, please squash some, though. :-)

bobmyhill:
Squished, squashed, and crushed...

Note to self: Never use "git merge --squash" :/

tjhei:
```git rebase -i master``` is your friend. :)

File: misc/ref/paper_fit_data.py.out [hunk 7]
```diff
@@ -1,9 +1,9 @@
-<stdin>:88: UserWarning: Overriding database equation of state. From BM2 to BM3
-<stdin>:97: UserWarning: Overriding database equation of state. From BM3 to BM2
+<stdin>:88: UserWarning: Warning, you are changing the method to BM3 even though the material is designed to be used with the method BM2.  This does not overwrite any mineral attributes other than temperature and pressure, which are set to nan. Stale attributes will be preserved unless they are refreshed (for example, by set_state).
+<stdin>:97: UserWarning: Warning, you are changing the method to BM2 even though the material is designed to be used with the method BM3.  This does not overwrite any mineral attributes other than temperature and pressure, which are set to nan. Stale attributes will be preserved unless they are refreshed (for example, by set_state).
 Optimization terminated successfully.
          Current function value: 14212.177931
-         Iterations: 117
-         Function evaluations: 255
+         Iterations: 116
+         Function evaluations: 246
```
ian-r-rose: Think I commented in the wrong place, sorry if this is a duplicate.

It seems like you are getting a different iteration count than my machine, though we are converging to the same values.  This line change makes the tests fail on my machine.  Are you able to reproduce this on master?

tjhei:
Can you catch the warning in test_endmembers.py like Ian does in test_composite.py?

misc/ref/paper_fit_data.py.out
tjhei: My machine is with Ian here. :-)

bobmyhill:
Done, and done. I'm becoming quite well-versed in squashing now :)

ian-r-rose:
Tests are working for me as well.  Thanks for putting up with all the nit-picking, Bob!  Safe to merge?

bobmyhill:
Safe to merge :)

PR #107, opened 2015-01-06
Fix prem

sannecottaar:
I cleaned out all the input files for prem, and fixed the values around the discontinuities in the mantle transition zone. I also tried to generalize SeismicRadiusTable, so it can handle tables with increasing depth or increasing radius (guess we could change the name SeismicRadiusTable then). If you guys are good with the file prem.txt, I will do the same for ak135 and iasp91.


burnman/data/input_seismic/PREMQL6.webarchive
```diff
@@ -1,509 +0,0 @@
-bplist00�_WebMainResource�
```
ian-r-rose: What is this file?  Is it necessary?

ian-r-rose:
I'm confused about all the different types of PREM.  Do they need to all be there, or can we merge them into one?

ian-r-rose:
Oops, egg on my face.  So much for reading comprehension... apologies, Sanne!

sannecottaar:
I could be a bit more elaborate. 
-- I replaced PRMQL6.webarchive, prem_lowermantle.txt, prem_table.txt, prem_table_nonSI.txt, prem_withQ.txt, grav_for_prem.txt with one file, prem.txt, wich includes Q values and gravity. 
-- I extended the example to plot these Q and gravity values.
-- I also changed my mind on seismologists always thinking along the radius, and also thought we should be able to accommodate files that have increasing depth (which prem.txt now is). The other option of increasing radius is tested by ak135 in the example. 
-- I think Fast() and Slow() could be removed, since we don't even want to use them in our inversion. But that aside.

sannecottaar:
Also included the option of depth bounds for internal_depth_list. This was important for a number of our examples (now that prem returns the entire earth and not just the lower mantle)

ian-r-rose:
So, it looks like the prem you have added here differs non-negligibly from what was there previously, at least in pressure (see, for example, example_grid.py).

sannecottaar:
Yeah, you are right that the pressure seems off at the CMB. I will improve
my integration. I'm afraid to say that the old version isn't necessarily a
better reference, as there were some weird jumps at the discontinuities,
not sure what happened there. If they were interpolated from somewhere
else, the values in the lower mantle might be fine.

ian-r-rose:
What is the latest on this, Sanne?

I just compared the old and new PREM models, and it seems that they are basically identical except in pressure.  As far as I can tell, the old pressures (as in, current master branch) are what is actually in Dziewonski and Anderson (1981).  

Where is an integration needed?  The original tables report all the information we need, as far as I can tell.

burnman/seismic.py
```diff
@@ -222,45 +280,59 @@ def v_s(self, depth):
...
     def density(self, depth):
 
         return self._lookup(depth, self.table_density)
 
     def depth(self, pressure):
-        if pressure > max(self.table_pressure) or pressure < min(self.table_pressure)  :
-           raise ValueError, "Pressure outside range of PREM"
-        radius = np.interp(pressure, self.table_pressure[::-1], self.table_radius[::-1] )
-        return self.earth_radius - radius
+        if self.table_depth[0]:
```
ian-r-rose: Should we no longer have this error check?  According to the documentation for interp, it returns the value on the edge of the model points if you request a point outside of the interval.

sannecottaar: Well, rebasing to master got me a new git trauma...

And test_partitioning.py is failing. I'm not sure why, but I have to some
other things now.

@Ian with regards to pressure: Yeah, I just reinterpolated pressure again.
I guess at the time I didn't really look at the mantle transition zone
being messed up (see attached figure) and the center of the Earth missing.
It still frustrates me that I cannot reproduce it at greater depths, even
when I use a more detailed table. And I currently don't have pressure for
ak135 and iasp91 (but I could look around).

@Ian with regards to error message: I think this error is mostly useful to
allude people to the fact they are using the wrong unit. Maybe it should
say that?

ian-r-rose: Oh, sorry, I wasn't clear. I agree this is a useful error message. It just looks like you are removing it.

sannecottaar: Oops...

sannecottaar: ok, the test_partitioning.py led to the depth routine in PREM being very
wrong (example_geotherms was broken too, but test.sh didn't catch that).
Otherwise test_partitioning looks wrong too, because it puts GPa into the
anderson geotherm instead of Pa. Not sure if the has anything to do with
you not reproducing it Bob.

sannecottaar:
I think this one is good to go. I took out some more bugs by testing examples individually. The test suite is working.

burnman/data/input_seismic/prem.txt
```diff
@@ -0,0 +1,95 @@
+#Depth(m)	Radius(m)	Pressure(Pa)	Gravity(m/s^2)	Density(g/m^3)	Vp(m/s)	Vs(m/s)	Qk	Qmu
```
tjhei: can we have a comment with the source here?

tjhei: also did you do any editing/computations from the source?

burnman/seismic.py
```diff
@@ -45,7 +45,7 @@ def evaluate_all_at(self, depth_list):
     
 
 
-    def internal_depth_list(self):
+    def internal_depth_list(self,mindepth=0.,maxdepth=6371.e3):
```
tjhei: What do these two arguments do? I don't think those are a good idea.

Please document.

burnman/seismic.py
```diff
@@ -155,21 +154,55 @@ def K(self, depth):
...
+        """
+        Parameters
+        ----------
+        depth : float or array of floats
+        Depth(s) [m] to evaluate seismic model at.
+
+        Returns
+        -------
+        Qk : float or array of floats
+        Quality factor for bulk modulus at given depth(s).
```
tjhei: does it have units?

burnman/seismic.py
```diff
@@ -155,21 +154,55 @@ def K(self, depth):
...
+        """
+        Parameters
+        ----------
+        depth : float or array of floats
+        Depth(s) [m] to evaluate seismic model at.
+        
+        Returns
+        -------
+        QG : float or array of floats
+        Quality factor for shear modulus at given depth(s).
```
tjhei: does it have units?

burnman/seismic.py
```diff
@@ -193,27 +226,39 @@ class SeismicRadiusTable(Seismic1DModel):
...
         
-    def internal_depth_list(self):
-        
-        return (self.earth_radius - self.table_radius)[::-1] #radius is sorted in increasing order, so we need to reverse the depth list
+    def internal_depth_list(self,mindepth=0., maxdepth=6371.e3):
+        if not len(self.table_depth) == 0:
+            return np.array([self.table_depth[x] for x in range(len(self.table_depth)) if self.table_depth[x]>mindepth and self.table_depth[x]<maxdepth]) 
+        else:
+            depth = self.earth_radius - self.table_radius[::-1]
+            return np.array([depth[x] for x in range(len(depth)) if depth[x]>mindepth and depth[x]<maxdepth])
```
tjhei: instead of creating this whole array every time this function is called, can you put the formula earth_radius-radius inside the ```[ ... ]```?

burnman/seismic.py
```diff
@@ -229,38 +282,46 @@ def density(self, depth):
...
     def depth(self, pressure):
         if pressure > max(self.table_pressure) or pressure < min(self.table_pressure)  :
-           raise ValueError, "Pressure outside range of PREM"
-        radius = np.interp(pressure, self.table_pressure[::-1], self.table_radius[::-1] )
-        return self.earth_radius - radius
+            raise ValueError, "Pressure outside range of SeismicRadiusTable"
+        if not len(self.table_depth) == 0:
+            depth = np.interp(pressure, self.table_pressure, self.table_depth )
+        else:
+            depth = np.interp(pressure, self.table_pressure[::-1], self.earth_radius - self.table_radius[::-1])
```
tjhei: please pass table_radius and then convert to depth after conversion.

burnman/seismic.py
```diff
@@ -229,38 +282,46 @@ def density(self, depth):
...
+        return depth
+
+    def radius(self, pressure):
+        if not len(self.table_radius) == 0:
+            radius = np.interp(pressure, self.table_pressure, self.table_radius )
+        else:
+            radius = np.interp(pressure, self.table_pressure[::-1], self.earth_radius - self.table_depth[::-1])
```
tjhei: same here

tjhei:
What is the reason for supporting radius and depth based tables in SeismicRadiusTable? It would be easier to convert depth to radius once while loading in the file. This would be faster and you need a lot less code.

sannecottaar:
I guess I attempted to keep the table the same as the input either
increasing radius or increasing depth. Yes, we could also flip the entire
table if needed.

sannecottaar:
There needs to be some kind of __init__ for every seismic model then, in case people define their own model.

burnman/seismic.py
sannecottaar: Not sure what you want me to change here?

burnman/seismic.py
sannecottaar: With this you can get the depth list for e.g. part of the PREM model instead of the entire earth.

burnman/seismic.py
tjhei: Not needed if you do everything in radius, but I was suggesting something like
```
   depth =  self.earth_radius - np.interp(pressure, self.table_pressure,self.table_radius)
```
so you don't create another temporary array

burnman/data/input_seismic/prem.txt
sannecottaar: I've calculated gravity, because although it is in the original paper, it isn't anywhere in the PREM data tables that I have. 
And values have been reinterpolated, and I gave discontinuities a finite width of 1km instead of 0 km, because I think various interpolations didn't deal with having two values for the same depth.

bobmyhill: I don't like giving discontinuities a finite width - PREM is no longer PREM in this case. Would it be better to have a line in the scripts that imposes a small dz (dr, dp) when z[i] = z[i+1]? Something like this, perhaps:
# Create interpolatable depths list
for i, depth in enumerate(depths):
    if depth <= depths[i-1]:
        depth += 1e-12

ian-r-rose:
I am still seeing differences of ~0.1-0.2 GPa in pressure, especially in the core:

compare_with_old_datafile

The x-axis is depth and the y axis is difference with the old datafile for prem.  The biggest changes are your fixes for the transition zone, clearly, but the wiggles in the rest of it seem odd to me.

Again, the old prem_table.txt appears to be a direct transcription from Dziewonski and Anderson.  Is there some update to the pressure profile that I'm unaware of?  To my eyes, the new one looks less smooth:
compre_data

sannecottaar:
Okay, this should be looking a lot better now. The pressure in PREM is fixed. 
Other changes are:
- adding a couple more 1D models (example_seismic.py is changed to show all these models)
- some of the models do not give gravity and pressure. I put in a simple cumtrapz integration to get these. They differ a bit from PREM, and that might be because there aren't enough layers in these models to this. Anything more precise, with spline fitting, I haven't gotten working yet (it is tricky with the discontinuities). However, I can open an issue and work on this later (as this pull request is getting bigger). 
- removing evaluate_all_at, as this was quite a rigid method (and one always had to look up the order of the output). Now you can 'evaulate' and give in a list of variables you want (although I guess one still has to look up the variable names). 

It would be helpfully if somebody else runs the tests, as a bunch of them are still failing for me. Thanks!

burnman/constants.py
```diff
@@ -19,3 +19,10 @@
 available in older versions.
 """
 Boltzmann = 1.3806488e-23 
+
+
+"""
+Newtonian constant of gravitation (G) in m^3 kg^-1 s^-2
+"""
+
+gravitational = scipy.constants.G
```
ian-r-rose: Not sure about this name--it's an adjective for a constant value. Perhaps Big_G, or just G, since it is namespaced by constants.py?

ian-r-rose:
A general question: I am looking at Dziewonski and Anderson right now, and there are still differences in pressure and gravity between Table II (pages 312 and 313) and the values in prem.txt.  Do you know what is going on there? Where is prem.txt from?

examples/example_build_planet.py
```diff
@@ -1,4 +1,4 @@
-# BurnMan - a lower mantle toolkit
+    # BurnMan - a lower mantle toolkit
```
ian-r-rose: Accidental indentation? That reminds me, we should update the copyright notice...

burnman/seismic.py
```diff
@@ -14,38 +21,30 @@ def __init__(self):
...
         Returns the lists of data for a Seismic1DModel for the depths provided
-            
+        
         Parameters
         ----------
+        vars_list : array of variables (in str format) to be evaluated
         depth_list : array of floats
-            Array of depths [m] to evaluate seismic model at.
-            
+        Array of depths [m] to evaluate seismic model at.
```
ian-r-rose: This, and other places: the numpydoc docstring format usually has an indentation in front of the description of a parameter. Not sure whether this affects the sphinx parsing.

sannecottaar:
I don't know why these values are given in the paper, but not in the
version published on the IRIS website.
Gravity is from the old grav_for_prem.txt. I think I computed this back at
Cider, but I forgot the details. The pressures are from the  PREM input
file to a normal mode code. So it has become quite a jumble. I will try to
hunt a more consistent file again.

Your other points are good, thanks!

ian-r-rose:
The IRIS-PREM model sounds pretty official as well, so if you think that is more appropriate, I have no objections... maybe a note to that effect?

misc/ref/debye.py.out
```diff
@@ -1 +1,2 @@
-error 7.013244e-14
+error 7.009307e-14
+time old 0.525255, time new 0.319653
```
ian-r-rose: I don't think anything should have changed here, and the error looks like machine precision, so there should be no reason to update these test results.

misc/ref/example_seismic.py.out
```diff
@@ -0,0 +1,3 @@
+pressure is not defined for IASP91
+gravity is not defined for IASP91
+density is not defined for IASP91
```
ian-r-rose: When I run the script I also get warning output about pressure and gravity for STW105 and AK135

sannecottaar: Yes, I put in warnings whenever it computes these values and they aren't
given. This is to warn if you put in your own, say upper mantle model, and
these calculations go totally wrong. I guess I could put in tests if it is
a whole earth model, but then again, we should maybe also add a moon model.

burnman/seismic.py
```diff
@@ -170,49 +201,77 @@ def depth(self, pressure):
...
         gravity : float or array of floats
-        Gravity for given depths in [m/s^2]
+            Gravity for given depths in [m/s^2]
         """
         raise ValueError, "not implemented"
         return -1
 
 
-class SeismicRadiusTable(Seismic1DModel):
+class SeismicTable(Seismic1DModel):
```
ian-r-rose: Any reason for the name change? Not that I object to it...

misc/ref/example_seismic.py.out
ian-r-rose: I think the warning is fine, but the test output is out of date with that
warning.

burnman/seismic.py
```diff
@@ -4,6 +4,13 @@
 
 import numpy as np
 import tools
+import constants
+import warnings
+import scipy.interpolate
```
ian-r-rose: This module is not currently used, until we attempt to do spline integration.

burnman/seismic.py
sannecottaar: Hmm, might be old. I think seven months ago, I tried to make it flexible
for both radius and depth, but have since taken that out again. But mostly
we search by depth anyways, so I think this shorter name is fine.

burnman/seismic.py
```diff
@@ -14,38 +20,31 @@ def __init__(self):
...
+    
+    def evaluate(self, vars_list, depth_list):
         """
         Returns the lists of data for a Seismic1DModel for the depths provided
-            
+        
         Parameters
         ----------
+        vars_list : array of str
+	    Variables to be evaluated. Options are depending on the model 'pressure','density','gravity','v_s','v_p','v_phi','G','K','QG','QK'
```
ian-r-rose: This was kind of difficult for me to read. Perhaps
```
Available variables depend on the seismic model, and can be chosen from ...
```

burnman/seismic.py
```diff
@@ -14,38 +20,31 @@ def __init__(self):
...
-        density = np.array([self.density(r) for r in depth_list])
-        v_p = np.array([self.v_p(r) for r in depth_list])
-        v_s = np.array([self.v_s(r) for r in depth_list])
-        v_phi = np.array([self.v_phi(r) for r in depth_list])
-        return pressures, density, v_p, v_s, v_phi
-    
+        
+        values=np.empty((len(vars_list),len(depth_list)))
+        for a in range(len(vars_list)):
+            values[a,:]=getattr(self,vars_list[a])(depth_list)
```
ian-r-rose: I like this change. Perhaps it would be a good idea to check whether a seismic model has this particular variable implemented? Of course, if it does not then an exception will be raised, so maybe that is enough.

ian-r-rose:
The output for example_compare_all_methods needs to be updated as well, though it may be difficult to do that from your end due to the difference in where warnings are piped... :(

examples/example_seismic.py
```diff
@@ -41,104 +41,78 @@
...
-    plt.plot(pressures3/1.e9,v_p3/1.e3,'-b', label='v_p fast')
 
-    plt.legend(loc='lower right')
-    plt.xlim([30,136])
-    plt.ylim([11,14])
-    plt.xlabel('pressure')
-    plt.ylabel('km/s')
+    
+    ## Alternatively one is able to evaluate all the variables for a model in a single line
+    #  values = model.evaluate(vars_list,depth_list)
```
ian-r-rose: This commented out line is unnecessary.

examples/example_seismic.py
```diff
@@ -41,104 +41,78 @@
...
+
+                # specify where we want to evaluate, here we map from pressure to depth
+                #1. format p = np.arange (starting pressure, ending pressure, pressure step) (in Pa)
+                #p = np.arange(1.0e9,360.0e9,1.e9)
+                #depths = np.array([models[m].depth(pr) for pr in p])
+                #2. we could also just specify some depth levels directly like this
+                #depths = np.arange(700e3,2800e3,100e3)
+                #3. we could also use the data points where the seismic model is specified over a depth range, this will bring out any discontinuities
+                #this is the preferred way to plot seismic discontinuities correctly
+                depths = models[m].internal_depth_list(mindepth=-1.e3, maxdepth=6372.1e3)
```
ian-r-rose: Why are we starting from a negative minimum depth? Perhaps a comment about that?

examples/example_seismic.py
```diff
@@ -41,104 +41,78 @@
...
-    #we could also use the data points where the seismic model is specified:
-    depths = s.internal_depth_list()
-
-    #now evaluate everything at the given depths levels (using interpolation)
-    pressures, density, v_p, v_s, v_phi = s.evaluate_all_at(depths)
+    #List of seismic 1D models 
+    models=[burnman.seismic.PREM(), burnman.seismic.STW105(),burnman.seismic.AK135(),burnman.seismic.IASP91()]
+    colors = ['r','b','m','k']
+    # Variables to plot
+    vars=['pressure','gravity','v_p','v_s','v_phi','density']#,
```
ian-r-rose: vars() is an internal python function which you are hiding here. Another name for this would be better

examples/example_seismic.py
```diff
@@ -41,104 +41,78 @@
...
+                # specify where we want to evaluate, here we map from pressure to depth
+                #1. format p = np.arange (starting pressure, ending pressure, pressure step) (in Pa)
+                #p = np.arange(1.0e9,360.0e9,1.e9)
+                #depths = np.array([models[m].depth(pr) for pr in p])
+                #2. we could also just specify some depth levels directly like this
+                #depths = np.arange(700e3,2800e3,100e3)
+                #3. we could also use the data points where the seismic model is specified over a depth range, this will bring out any discontinuities
+                #this is the preferred way to plot seismic discontinuities correctly
+                depths = models[m].internal_depth_list(mindepth=-1.e3, maxdepth=6372.1e3)
+                #now evaluate everything at the given depths levels (using interpolation)
```
ian-r-rose: using interpolation --> using linear interpolation

examples/example_seismic.py
```diff
@@ -41,104 +41,78 @@
...
+                try:
```
ian-r-rose: This try/except block is very economical, but maybe not clear to somebody who doesn't know what is going on.  I think some more comments are in order about what you are doing here.

burnman/seismic.py
```diff
@@ -170,49 +200,77 @@ def depth(self, pressure):
...
+        #Shift values at discontinities by 1 m to simplify evaluating values around these.
+        depths[discontinuities]=depths[discontinuities]-1.
+        depths[discontinuities+1]=depths[discontinuities+1]+1.
+        return depths
+
 
     def pressure(self, depth):
-        
-        return self._lookup(depth, self.table_pressure)
+        if len(self.table_pressure)>0:
```
tjhei: you can simplify this logic to say:
```
if len(bla)==0:
  compute()
return lookup(bla)
```

burnman/seismic.py
```diff
@@ -170,49 +200,77 @@ def depth(self, pressure):
...
+            return self._lookup(depth, self.table_pressure)
+        else:
+            if len(self.table_density)>0:
+                warnings.warn("Pressure is not given in "+ self.__class__.__name__ +" and is now being computed. Potentially this might not be very realistic if "+ self.__class__.__name__ +" is not a global model. Potentially it might be better to use the depth/pressure conversion given by the PREM model.")
+
+                self._compute_pressure()
+                return self._lookup(depth, self.table_pressure)
+
+    def gravity(self, depth):
+        if len(self.table_gravity)>0:
```
tjhei: same here

examples/example_seismic.py
tjhei: also remove the comment behind it

burnman/seismic.py
```diff
@@ -14,38 +20,31 @@ def __init__(self):
...
-    
+        
+        values=np.empty((len(vars_list),len(depth_list)))
+        for a in range(len(vars_list)):
+            values[a,:]=getattr(self,vars_list[a])(depth_list)
+        return values
 
 
-    def internal_depth_list(self):
+    def internal_depth_list(self,mindepth=0.,maxdepth=6371.e3):
```
tjhei: do we need the defaults here?

tjhei:
Looks pretty good otherwise (if tests are working)

burnman/seismic.py
sannecottaar: Why not? Now it still works in the old way where you just get the depths in the list (at least if it is smaller than Earth).

burnman/seismic.py
sannecottaar: Yeah, the attributeerror is quite clear.

sannecottaar:
I removed the gravity from the PREM file, and the values are now computed
(and are a lot closer to the values in the PREM paper). I guess that is
more honest, since I remember computing them before, and they are not from
some reference file.  The pressure values are not exact, but they are close
enough (I think).

sannecottaar:
Thanks guys for all the feedback. 
Ian, would you run the tests? Indeed I have trouble dealing with the new warnings, maybe somebody can fix that right after merging this. But other tests are also failing as I don't have some text_2path or numdiff. Thanks!

burnman/seismic.py
tjhei: Which means you are hardcoding this for Earth. Couldn't you put in something like 1e100 for ```maxdepth```?

ian-r-rose:
For what it's worth, I also have a `text_2path` related error, but I thought it was just me. I think it is due to a newer version of matplotlib not working for the `Template` backend. This is probably a bug in matplotlib.

Sanne, can you try running the test script after commenting out line 21? You will have to close a bunch of figures, but you may see the `text_2path` error disappear.

ian-r-rose:
Is the warning about calculating pressure and gravity still accurate? Also, I think the wording of the warning is kind of unclear regardless.

ian-r-rose:
Okay @sannecottaar, when this is ready, I can update the test results.

sannecottaar:
Updated the wording of the warning. Can we merge?

tjhei:
What about the max_depth default value for |internal_depth_list| in the base class? I think the default there should also be a large value instead of Earth's depth. The documentation could also be extended to explain these min/max values.

sannecottaar:
Sorry, had only fixed this in SeismicTable.

tjhei:
Looks good to me now. @ian-r-rose ?

ian-r-rose:
@sannecottaar, can you cherry-pick this test-results update from my repo? 
ian-r-rose@0b59eae7f2917bf7ef19cd13f479b164ec846f36

ian-r-rose:
Okay, looks good from my end. Someday it may be nice to revisit what is going on with the pressure and gravity profiles in PREM, but I don't think it is fruitful at the moment.

PR #108, opened 2015-01-07
modernizing examples

sannecottaar:
Some examples were broken, and in others I just put a more 'modern' rock. I'm aware that this will probably break a bunch of tests as outputs might have changed, but I don't know how to fix this.

### DIFF
--- a/examples/example_averaging.py
+++ b/examples/example_averaging.py
@@ -56,12 +56,12 @@ if __name__ == "__main__":
     amount_perovskite = 0.6
 
     rock = burnman.Composite([amount_perovskite, 1.0-amount_perovskite],
-                             [minerals.SLB_2005.mg_perovskite(),
-                              minerals.SLB_2005.periclase()])
+                             [minerals.SLB_2011.mg_perovskite(),
+                              minerals.SLB_2011.periclase()])
 
-    perovskitite = minerals.SLB_2005.mg_perovskite()
+    perovskitite = minerals.SLB_2011.mg_perovskite()
 
-    periclasite = minerals.SLB_2005.periclase()
+    periclasite = minerals.SLB_2011.periclase()
 
     #seismic model for comparison:
     # pick from .prem() .slow() .fast() (see burnman/seismic.py)
--- a/examples/example_composition.py
+++ b/examples/example_composition.py
@@ -15,18 +15,10 @@ composition. Here we present a couple of examples:
 
 1. Two minerals mixed in simple mole fractions. Can be chosen from the BurnMan
    libraries or from user defined minerals (see example_user_input_material)
-2. Two minerals mixed in simple mole fractions with user-defined Fe
-   partitioning
-3. The user can input wt% of each cation (Mg, Fe and Si) and BurnMan will
-   calculate Fe partitioning along a :math:`P, T` profile (see
-   example_partition_coef.py)
-4. A mixture of three minerals.
+2. Example with three minerals
+3. Using preset solid solutions
+4. Defining your own solid solution
 
-In compositions 2, 3, and 4 of the above inputs, BurnMan will mix the mineral
-physical paremeters of end member minerals (pure Mg and Fe) of the user's
-choice using either volumetric (moduli) or molar averaging (all others) at
-room pressure and temperature (see example_user_input_material.py for
-information on these parameters).
 
 To turn a method of mineral creation "on" the first if statement above the
 method must be set to True, with all others set to False.
@@ -38,11 +30,13 @@ example_spintransition.py for explanation of how to implement this
 
 * :doc:`mineral_database`
 * :class:`burnman.composite.Composite`
-
+* :class:`burnman.minerals.Mineral`
+* :class:`burnman.solidsolution.SolidSolution`
 
 *Demonstrates:*
 
 * Different ways to define a composite
+* Using minerals and solid solutions
 * Compare computations to seismic models
 
 """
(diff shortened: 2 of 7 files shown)


sannecottaar:
P.s. are we set on 'base_material' in SolidSolution, or could we still name this 'endmembers', since we also have 'n_endmembers'.

bobmyhill:
I would be very happy to call them endmembers. "for endmember in endmembers" is much more sensible than "for endmember in base_material"

ian-r-rose:
I am happy with endmembers.  Someone just has to do it :)

bobmyhill:
I'm happy to do it... give me a few minutes...

bobmyhill:
Ok, pull request opened...

ian-r-rose:
Sanne, this branch is out of date, which is making it tough to run the tests.  Can you rebase it onto master?

sannecottaar:
Done. But I expect a lot of issues with the tests, as a lot of the output
changed.

bobmyhill:
Do you want me to check/change the test output, or are you happy to do this? It's pretty simple ... all the comparison files are in misc/ref - they need changing after checking that nothing has gone wrong...

ian-r-rose:
The tester just checks the output of the script against some reference previous output in misc/ref.  In order to update the output, all you need to do is copy the new output into that folder, with the appropriate name.

sannecottaar:
Yes, that looks doable. If it wasn't that the test suite is now already
failing at test_partitioning.py, which I think should not have changed.

ian-r-rose:
I think the test_partitioning.py failure is in your other open PR.  This one looks fine to me.

ian-r-rose:
Sanne, what is the status of this?  Is there a merge commit for the base_materials->endmembers change here?

sannecottaar:
Ok, I changed some of the output files for the examples. I don't know why the comparison with master shows all the base_materials to endmembers changes, since that came from merging with master.

I do find the testing of the examples a bit random. Do we really want this? The original output to screen is a bit random in the first place. I was able to change some examples without breaking them. On the other hand example_composition.py will be broken if one of the other (True/False)-options gets chosen.

bobmyhill:
"I don't know why the comparison with master shows all the base_materials to endmembers changes, since that came from merging with master."

That's a "feature" of merging. I think that in general *git rebase -i* master is much cleaner than git merge.
As the merge wasn't a separate commit, we can't undo it easily.

Timo and I discussed the example tests in Berkeley - they are obviously not thorough tests of the examples, but they do help to identify when we've done something badly wrong.

sannecottaar:
Yeah, I guess it does show when errors are popping up.
I will try 'git rebase -i master' next time.

tjhei:
if you rebase to master now all the ```base_materials``` to ```endmembers``` changes should go away.

sannecottaar:
It claims it is up to date.

tjhei:
Nope. Switch to master and update it first:
```
   git checkout master
   git pull --ff-only upstream
   git checkout update_examples
   git rebase master
```

bobmyhill:
Hmmm, test_partitioning is rather opaque to me. I tried to check your old and new a and b values using eq. 6 from Nakajima et al., 2012, and neither of them agree with my calculations. It might be a problem with my calculations, though...

Are a and b  X_per_FeO/X_per_MgO and X_pv_FeSiO3/X_pv_MgSiO3 respectively?



The additional imports for test_partitioning.py are
import numpy as np
import burnman.constants

The check lines I added were these ones, which should sum to zero:
        a = 0.0319126180713
        b = 0.125395826369
        X_per_FeO = a/(a+1)
        X_pv_FeSiO3 = b/(b+1)
        print burnman.constants.gas_constant*temperature*np.log((X_pv_FeSiO3 / (1-X_pv_FeSiO3)) / (X_per_FeO/(1.-X_per_FeO))) - (-22300. - 200.*(pressure/1.e9) - 4.0*temperature + 4900.*(1. - 2.*X_pv_FeSiO3) + (11000. + 110.*(pressure/1.e9))*(1. - 2.*X_per_FeO))

ian-r-rose:
I think this particular test should be on the chopping block. That being said, it should not be broken by these PRs.

bobmyhill:
We definitely need a test for the functions in partitioning.py though, especially as I can't reproduce the values...

sannecottaar:
The problem with the partitioning example is only in fix_prem. As far as
I'm aware, this one can be merged.

tjhei:
tests pass here

bobmyhill:
Can you remove the stray floats in the mineral definitions? I don't think they do anything.
e.g. minerals.SLB_2005.mg_fe_perovskite(0.7) -> minerals.SLB_2005.mg_fe_perovskite()

On a related note, I was considering modifying solid solution instances to allow lists of floats to set_composition on init, like this:
pv_composition=[0.9,0.1,0.0]
pv=minerals.SLB_2005.mg_fe_perovskite(pv_composition)

Not for this pull request, of course, but what do you think?

bobmyhill:
There are some more stray fe_nums in 
* examples/example_geotherms.py and  
* examples/example_woutput.py

sannecottaar:
I think I have already removed those. At the moment the fe_num still seem
correct for the 2005 database. Should we change that? I updated the
examples to the 2011 database, and included set_composition statements. I
agree, including set_composition in init is nice.

bobmyhill:
Ah, yes, I was forgetting the differences with the 2005 database. Eventually we can change it to be completely consistent with the 2011 database, but that's not for this PR, I think.

Are you happy with this one now? If so, I'll merge.

sannecottaar:
Yep.

PR #109, opened 2015-01-07
rename base_material to endmembers

bobmyhill:
Renames *base_material* and *base_materials* to *endmembers*.


ian-r-rose:
Looks good to me.

PR #110, opened 2015-01-08
Add new example for building a planet

ian-r-rose:
I promised this one a while back.  This is an example which self-consistently calculates pressure, gravity, and density for a planet (using Mercury as an example).  It then calculates the planet's mass and moment-of-inertia, which can be compared to geodetic constraints.  Kind of a neat example of the kind of things we can do...


tjhei:
Cool stuff. Can you list this in test.sh and add the reference comparison file?

ian-r-rose:
Added.

tjhei:
great. Anybody else have any science comments?

lordkman:
Looks good to me. One bug though is if you run it from your initial burnman
directory (not cd'ing into examples) you get:

"-bash: cd: examples/example_build_planet.py: Not a directory"

Also, as an example, perhaps put things like mass and radius in relative
units? Makes changes from known planet data a bit more obvious than ~10 sig
figs of output on the command line.

-Cayman

ian-r-rose:
I don't know how that error is coming up, but in general none of the examples work except when run from the same directory as the scripts.

I did think about putting the mass in some relative units, but I could not think of anything good to compare against.  Suggestions? Moment of inertia is normalized to by MR^2, so at least there is a natural unit for that...

ian-r-rose:
I suppose I could compare with the actual observed mass of Mercury... which is not very close to the result of the simple model.

lordkman:
Running the calculation it looks like you used a radius cutoff but the mass
is 25% too large. My guess then is that either the core-radius is too big
or something is up with the Fe.

As for moment of inertia, I'd just go with MR^2 to play it safe although I
have zero experience with using that as a datum.

ian-r-rose:
I am not actually trying to fit the planet in this example, as it would involve a much more complicated model for the interior.  This is a simple two-layer model with a solid iron core.  Yes, the resulting mass is too large by about 35%.

lordkman:
Well you've got me curious as to what's wrong. While Mercury is weird you
generally can fit things pretty well with a simple layer cake model. Was
the CMB radii just a guess?

Should we benchmark this to say, Earth? I don't mind doing that. My M-R
calculations of just Fe and Pv gets the Earth-radii within 10% and I'd be
curious to compare

-Cayman

ian-r-rose:
There are three things that are completely unrealistic here.  First, the core is solid, when we know there is at least some liquid in real life.  Second, there is no alloying element in the core, I've used pure iron.  Third, the temperature profile is bogus.  A more complex model would do a much better, I'm sure.

bobmyhill:
Very cool :)

Why are you using perovskite and periclase?! I know burnman used to be a lower mantle toolkit, but using dense high pressure minerals to model a small planet seems a bit eccentric :D 

Why not use olivine and pyroxene, now that we have solution models for both of them? I'm pretty sure that will get you close to Mercury mass.

ian-r-rose:
That is a good point!  Fixed, though the bigger issue for matching the mass of the planet is the core, so this is only a marginal improvement.

lordkman:
I've been playing around with this and scaling it up to the Earth. But in
looking at it I can't quite tell if the mass and pressure are coupled ODE's
rather than two ODE's calculated separately. Does this just evoke a
pressure field, calculate the density and thus mass given the pressure
field and go from there? Because as far as I can tell you shouldn't need
multiple iterations at all. Solve the mass within a shell and hydrostatic
EQ simultaneously with the EOS providing rho as a function of P (or vice
versa). So you'd evoke a central pressure and some dr and integrate by
shells outward using a mass or radius cutoff to stop the integration.
Thoughts?

-Cayman

ian-r-rose:
Well, in reality the density, gravity, and pressure equations are all coupled differential equations.  This coupling makes it a highly nonlinear system, though.  By turning it into a fixed-point iteration as I have done, we can integrate it as a series of ODEs very simply.

My question for you would be: from where do we get the central pressure and gravity profile?

lordkman:
I've never calculated this with gravity, instead using mass and pressure,
so I can't say anything to that.

P_cen is sadly a guess. This is where the mass or radius cutoff comes in.
Ideally (say for the Earth), you'd 1 earth radius and 1 earth mass, but a
simple two layer model doesn't typically give that. So you can either tweak
P_cen or add more compositional complexity.

The equations I and most of the mass-radius literature solve are:
mass within a shell:
dm/dr = 4*pi*r^2*rho(r)

hydrostatic EQ:
dP/dr = -G*m(r)*rho(r)/r^2

and EOS:

P(r) = EOS(rho(r),T(r))

So initial conditions are dependent on P_cen and dr. Create a sphere of
radius dr with constant pressure P_cen, solve the coupled ODES to determine
dP and dm within this shell, go to the next dr and repeat until you a mass
or radius cutoff. These are fairly well behaved functions even across the
CMB (I use a mass cutoff to distinguish between C and M for "Earth").

lordkman:
I found that switching hydrostatic EQ from Rho*g*z to the equation I use
with m(r) and rho(r) caused the central pressure to increase exponentially
with each iteration until basically the EOS broke down. I think Mercury was
okay because its maximum pressure was small compared to the Earth.

ian-r-rose:
Hmmm.  I don't quite see at the moment where your form of the hydrostatic equation comes from, but why switch it from rho g z? I think that the script should work just fine for Earth-like values.

As it is, note that I am integrating the hydrostatic equation downwards from the surface (at least we know the pressure there :) ), so that might be a difference you would need to address.

lordkman:
We're describing the same equation, mine is in terms of mass, yours in in
terms of g. And I've just always prefered to do things in terms of r rather
than depth because I don't like switching between bottom -> top and then
top - > bottom for reasons of making sure I'm consistent.

The code though doesn't seem to calculate pressure self-consistently (as
coupled ODE's). It guesses at P, then calculates the densities and gravity
given that profile, then recalculates pressure. Why recalculate pressure at
all?

ian-r-rose:
Yes, but what you have described involves making the correct guess for the center pressure.  If you want to get an answer that matches (radius, mass) you will need to make a bunch of guesses at that pressure.  I'll bet that the number of guesses you need to make are at least as many as the Picard iterations.

It seems to me that integrating the hydrostatic equation from the center like that will ultimately be a sort of iterative-by-hand approach.

lordkman:
I think that's how you have to do it though. Guess and check on P_cen until
you get the right mass and radius. I can't see any other way of doing it
and remaining consistent. You can say for sure that P = 0 at whatever
radius maximum radius given this model and be consistent. Perhaps just say
the planet is built when P at a given r is 0 regardless of M or R.

From what I can tell it's doing right now is pinning the radius (and thus
mass) of/at the surface and trying to match a pressure (and thus density)
profile to fit it. If you further pin the radius of the CMB, this seems a
very likely to break (and does for Earth-like pressures).

bobmyhill:
A couple more thoughts:
* Can any of these functions be farmed out to a generalised module, so that people can define their own planet more easily? 
* From your previous conversation, it seems that top-down and bottom-up both have their pros and cons. I would have calculated properties from the top down, like Ian has done, but if both are used, then I see no reason not to implement both.
* How difficult would it be to add the Adams-Williamson equation here? Since it's used a lot in practicals, and it's inherently cool, it might be a nice function to add...

All looks good to me, and I'd be happy to merge if you think the above can wait until a later date.

ian-r-rose:
Yes, in fact adding them to a general module is on my longer-term todo
list.  One step at a time, though.

No reason not to do both top-down and bottom-up approaches. though I would
table that for when I make a more general module.

It shouldn't be too hard to implement the Adams-Williamson equation either,
though I would need to then treat the temperature profile more carefully.
As it is I am more or less ignoring it.  In general, you would assume a
constant composition, adiabatic profile, and then have some prescribed
temperature jumps at the boundaries representing thermal boundary layers.
So that would add some more complexity, as well as N new parameters, where
N is the number of convecting layers.

To summarize, yes, these are all good ideas, but I don't think I'm going to
address them right away in this relatively simple example :)

I don't have strong opinions whether to merge now or later -- I think the
priority should be to tie up some loose ends for the current release
(mostly resolve Sanne's outstanding PRs and fix ipython-notebooks)

bobmyhill:
Yes, I'm happy for my suggestions to wait.
As for the notebooks. Is anyone doing this? If not, I can do it.
For the current release, I would also like to include my outstanding PR - if no-one has any complaints, that is. I have a few assemblages manually defined in the script for a manuscript I'm currently writing, and it would be much neater if I could use the new class and cite the current release rather than a private branch...

bobmyhill:
Ok, I'm happy with this one... it's merging time!

PR #111, opened 2015-01-10
Assemblage Class

bobmyhill:
The Assemblage class is similar to the Composite class, but takes only a list of phases as input. 
* Following the structure in the Composite class, phases can be nested arbitrarily.
* Following the code structure in the Mineral and SolidSolution classes, it is possible to set_phase_fractions, set_composition for each phase and set_state.
* set_state has been split according to an optional parameter, "none", "gibbs_only", "elastic", "thermodynamic", "all", which compute different properties of the assemblage.
* Mineral, SolidSolution and Assemblage now have composition as an attribute
* Phase fractions can be defined as "molar", "mass" or "volume". "molar" is the standard (an atomic balance constraint, and the most sensible from a thermodynamic perspective), and molar_fractions is always set as an attribute of the assemblage class when set_phase_fractions is set.
* Velocity calculations are based on volume fractions, so molar_fractions are converted when calculate_moduli is called.
* The chemical potentials / fugacities of an assemblage can also be found trivially.

### DIFF
--- a/burnman/assemblage.py
+++ b/burnman/assemblage.py
@@ -200,4 +200,4 @@ class Assemblage(Material):
         mat_rho = np.array([m.rho for m in moduli])
         mat_K = np.array([m.K for m in moduli])
         mat_G = np.array([m.G for m in moduli])
-        return mat_rho, mat_vp, mat_vs, mat_vphi, mat_K, mat_G
+        return mat_rho[0], mat_vp[0], mat_vs[0], mat_vphi[0], mat_K[0], mat_G[0]
--- a/burnman/assemblage.py
+++ b/burnman/assemblage.py
@@ -45,7 +45,8 @@ class Assemblage(Material):
         phases: list of :class:`burnman.Material`
             list of phases.
         """
-        assert(len(phases)>0)
+        if len(phases)==0:
+            warnings.warn("Assemblage must be populated with phases before you can do anything with it!", stacklevel=2)
         self.phases=phases
 
 
@@ -54,7 +55,7 @@ class Assemblage(Material):
         indent += "  "
         for (fraction, phase) in self.children:
             print "%s%g of" % (indent, fraction)
-            phase.debug_print(indent + "  ")
+            phase.debug_print(indent + "  ")	
 
     def set_phase_fractions(self, fractions):
         assert(len(self.phases)==len(fractions))
@@ -65,7 +66,7 @@ class Assemblage(Material):
         self.fractions = [max(0.0, fr) for fr in fractions]  # turn -1e-14 into 0.0
         total = sum(self.fractions)
         if abs(total - 1.0) > 1e-12:
-            warnings.warn("Warning: list of molar fractions sums to %g. Normalizing." % total)
+            warnings.warn("List of molar fractions sums to %g. Normalizing." % total, stacklevel=2)
             self.fractions = [fr / total for fr in self.fractions]
 
     def set_composition(self, phase_compositions):
--- a/burnman/assemblage.py
+++ b/burnman/assemblage.py
@@ -57,17 +57,52 @@ class Assemblage(Material):
             print "%s%g of" % (indent, fraction)
             phase.debug_print(indent + "  ")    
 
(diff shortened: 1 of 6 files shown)


bobmyhill:
To do:
* We should probably have the option to define molar, mass and volume fractions. 
* As I've ripped rho / V from different places, I need to check that these are correct.
* There should be an option to output the bulk composition of the assemblage if all of the phases has an associated chemical formula.

bobmyhill:
Ok, so I've dealt with all the problems I found with this class. 
I think it's a natural extension to the classes we have already, and it makes a number of things easy to do. I'd be interested to hear your opinions...

If you like the way I've split set_state, maybe we can do something similar for the Mineral and SolidSolution classes...

ian-r-rose:
I think there is a lot of good stuff here, but I would advocate holding off on this PR for the current release for a few reasons:

* This is a big addition -- maybe too big for 0.8?
* There is clearly (as you have noted) a lot of similarities with the Composite class.  In fact, a number of the things you have done here are the same as things that I have wanted to add to Composite (such as being able to talk to an AveragingScheme).  I am not completely convinced that we shouldn't just augment Composite with these new functionalities.
* Timo expressed some reservations about having Composite be able to update its phases and fractions.  I'm not sure I agree -- those functions which you have implemented here might be a fine fit in Composite.
* I do like the way you have split up set_state.  However, this is somewhat in conflict with the way that Timo has been trying out in the performance optimizations branch.  I think that we would want to do the "conditional calculation of various properties" in a unified way.

bobmyhill:
Ok, to reply to your points:
* It is a big addition, true. It's nothing groundbreaking though - it just makes it a lot easier to calculate properties of assemblages.
* If are convinced that Composite should take fractions at initialisation, a separate Assemblage class is necessary. There are functions in Assemblage that don't require phase fractions to be set (specifically the chemical potential functions), and I quite like these being included in one place. 
* I agree that we should have one consistent way to split set_state.

burnman/assemblage.py
```diff
@@ -0,0 +1,312 @@
...
+            for element in composition:
+                if element not in self.composition:
+                    self.composition[element] = self.molar_fractions[i]*composition[element]
+                else:
+                    self.composition[element] += self.molar_fractions[i]*composition[element]
+        return self.composition
+
+    def set_method(self, method):
+        """
+        set the same equation of state method for all the phases in the composite
```
tjhei: you should call base class set_method

tjhei:
I don't have a lot of time right now, but so that I don't forget:
- I think we should not change ```set_state``` this way but do it from ```Material``` down and consistently.
- some functions look like they should be internal
- most of the functionality could go into ```Composite``` (like changing the fractions) but I am unsure on where to draw the line
- if we allow the user to access the phases, we wouldn't need ```set_composition```?

burnman/assemblage.py
bobmyhill: Ok, I'm being slow here ... base class set_method? How do you mean?

bobmyhill:
*I think we should not change set_state this way but do it from Material down and consistently.
** Ok, I'll need a little help understanding what that means! I agree that we should do things consistently, but I thought this was pretty consistent. Do you mean that we should unroll and set state? 

*some functions look like they should be internal
** ok, this can be changed easily. I think I know which ones you mean.

*most of the functionality could go into Composite (like changing the fractions) but I am unsure on where to draw the line
** Ok, well, the functionality in Assemblage right now is pretty common stuff for thermodynamicists, so I'm happy for a) everything to be shifted to Composite, or b) keeping this class. If (a), I really want to be able to just initialise the assemblage (without fractions). 

*if we allow the user to access the phases, we wouldn't need set_composition?
** True. This was a brevity and completeness issue - there may be another way to solve the problem. The logic behind me designing it like this is twofold:
*** The only variables normally fixed in thermodynamics are the bulk composition constraints. It seems sensible to allow the user to vary phase fractions and compositions at the same time.
*** I don't like the idea of having a six-phase assemblage and needing eight (or more) lines of code just to change the composition. I know it aids legibility though.

PR #113, opened 2015-01-11
test.sh: make numdiff optional, make run on MAC

tjhei:
readlink is not supported on mac, but it turns out we don't need it.
Make numdiff optional. Tests will run but some tests will fail using
diff.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -2,6 +2,16 @@
 
 fulldir=`pwd`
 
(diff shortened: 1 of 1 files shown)


tjhei:
@sannecottaar can you check if this works for you now?

sannecottaar:
Yes, great! That solves the biggest error I come across in test.sh...

PR #114, opened 2015-01-11
do not use grep -P

tjhei:
This might make it work on MAC.


tjhei:
@sannecottaar does this fix the grep error?

sannecottaar:
Yes, another error bites the dust. Thanks, Timo!

PR #117, opened 2015-01-14
Molar fractions and state passed to SolidSolution

bobmyhill:
The optional argument *molar_fraction* is added to the __init__ of the SolidSolution class. This allows the user to set the [composition] of the solid solution with a single line rather than two. The order of variables is chosen for compatibility with older SolidSolution classes.

### DIFF
--- a/burnman/solidsolution.py
+++ b/burnman/solidsolution.py
@@ -27,7 +27,7 @@ class SolidSolution(Mineral):
     and P derivatives in J/K/mol and m^3/mol.
     """
 
-    def __init__(self, endmembers, solution_model=SolutionModel()):
+    def __init__(self, endmembers, solution_model=SolutionModel(), molar_fraction=None):
         """
         Set up matrices to speed up calculations for when P, T, X is defined.
 
@@ -48,6 +48,10 @@ class SolidSolution(Mineral):
 
         for i in range(self.n_endmembers):
             self.endmembers[i][0].set_method(self.endmembers[i][0].params['equation_of_state'])
+
+        if molar_fraction is not None:
+            self.molar_fraction = molar_fraction
+            self.set_composition(molar_fraction)
 
     def get_endmembers(self):
         return self.endmembers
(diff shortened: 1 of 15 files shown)


bobmyhill:
What do you think? Especially @tjhei as our arbiter of syntax :)
I split the commits relevant to molar fraction and those relevant to state, just in case you didn't like the latter. I think this will make for much more compact code. I was considering doing the same for the mineral class (including [P,T] as an __init__ argument), but decided to wait to see what you thought of this first.

ian-r-rose:
I don't have super strong feelings about it.
Overall, I would say that putting *molar fraction* in the initializer makes good sense, but I am less enamored with putting the P,T state in the initializer, as this is something that is likely to change all the time, and feels less appropriate to it.

bobmyhill:
Yup, I'm in two minds about the state part as well. I'm happy to remove this part ... it didn't take long to write.

tjhei:
I agree with Ian, what is the reason to include the state?

bobmyhill:
Ok, so as I was ambivalent about adding state (I added it for code brevity only), and you're both against including it, I've removed those commits from the history. This should be ready to roll.

burnman/solidsolution.py
```diff
@@ -27,7 +27,7 @@ class SolidSolution(Mineral):
     and P derivatives in J/K/mol and m^3/mol.
     """
 
-    def __init__(self, endmembers, solution_model=SolutionModel()):
+    def __init__(self, endmembers, solution_model=SolutionModel(), molar_fraction=None):
```
ian-r-rose: This seems like a bug.  In some places you are calling it molar_fraction, and in others you are calling it molar_fractions

bobmyhill: Not a bug (I thing I've been consistent in each file). It's just a horrible legacy from our mega commit in December. We deleted the commit where I fixed this. I can change them all to molar_fractions...

ian-r-rose: Ah, I see now.  My mistake.

bobmyhill: Fixed anyway ... I meant to (re)do this a while ago :)

bobmyhill:
Can this be merged now?

tjhei:
Sorry for the delay. Looks good, tests pass.

PR #121, opened 2015-03-01
Magnetic contribution

bobmyhill:
Adds terms accounting for the magnetic contribution to the Gibbs Free Energy. Important for phases like body centered cubic (BCC) iron.

### DIFF
--- a/burnman/eos/hp.py
+++ b/burnman/eos/hp.py
@@ -282,10 +282,11 @@ class HP_TMT(eos.EquationOfState):
         Returns the magnetic contribution to the Gibbs free energy
         """
 
-        structural_parameter=params.magnetic_structural_parameter
+        structural_parameter=params['magnetic_structural_parameter']
+        tau=temperature/(params['curie_temperature'][0] + pressure*params['curie_temperature'][1])
+        magnetic_moment=params['magnetic_moment'][0] + pressure*params['magnetic_moment'][1]
+
         A = (518./1125.) + (11692./15975.)*((1./structural_parameter) - 1.)
-        tau=temperature/(params.curie_temperature[0] + pressure*params.curie_temperature[1])
-        magnetic_moment=params.magnetic_moment[0] + pressure*params.magnetic_moment[1]
         if tau < 1: 
             f=1.-(1./A)*(79./(140.*structural_parameter*tau) + (474./497.)*(1./structural_parameter - 1.)*(np.power(tau, 3.)/6. + np.power(tau, 9.)/135. + np.power(tau, 15.)/600.))
         else:
--- a/burnman/eos/hp.py
+++ b/burnman/eos/hp.py
@@ -279,7 +279,9 @@ class HP_TMT(eos.EquationOfState):
 
     def _magnetic_gibbs(self, pressure, temperature, params):
         """
-        Returns the magnetic contribution to the Gibbs free energy
+        Returns the magnetic contribution to the Gibbs free energy [J/mol]
+        Expressions are those used by Chin, Hertzman and Sundman (1987)
+        as reported in Sundman in the Journal of Phase Equilibria (1991)
         """
 
         structural_parameter=params['magnetic_structural_parameter']
--- a/burnman/eos/hp.py
+++ b/burnman/eos/hp.py
@@ -153,7 +153,12 @@ class HP_TMT(eos.EquationOfState):
             else:
                 Gdisord=0.0
 
-        return params['H_0'] + self.__intCpdT(temperature, params) - temperature*(params['S_0'] + self.__intCpoverTdT(temperature, params)) + intVdP + Gdisord
+        if params.has_key('magnetic_moment'):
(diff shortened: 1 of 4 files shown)


tjhei:
Great.

1. Can somebody who has some geochemistry knowledge look at it?
2. Could you please add a simple unit test that computes this term?

burnman/eos/hp.py
```diff
@@ -272,7 +277,22 @@ def __intCpoverTdT (self, temperature, params):
         """
         return (params['Cp'][0]*np.log(temperature) + params['Cp'][1]*temperature - 0.5*params['Cp'][2]/np.power(temperature,2.) - 2.0*params['Cp'][3]/np.sqrt(temperature)) - (params['Cp'][0]*np.log(T_0) + params['Cp'][1]*T_0 - 0.5*params['Cp'][2]/(T_0*T_0) - 2.0*params['Cp'][3]/np.sqrt(T_0))
 
+    def _magnetic_gibbs(self, pressure, temperature, params):
+        """
+        Returns the magnetic contribution to the Gibbs free energy
```
sannecottaar: Holy guacamole, these equations baffle me. How do people get to these ratios? As far as I can follow it, it looks good. Could you add the Sundman reference here?

burnman/minerals/Sundman_1991.py
```diff
@@ -0,0 +1,71 @@
...
+            'V_0': 6.93863394593e-06 ,
+            'Cp': [22.24, 0.0088656, -221517., 47.1998] ,
+            'a_0': 5.13e-05 ,
+            'K_0': 1.539e+11 ,
+            'Kprime_0': 5.2 ,
+            'Kdprime_0': -3.37e-11 ,
+            'n': sum(formula.values()),
+            'molar_mass': formula_mass(formula, atomic_masses),
+            'curie_temperature': [201., 0.0] ,
+            'magnetic_moment': [2.1, 0.0] ,
```
sannecottaar: Are Tc and beta negative for fcc?

bobmyhill: There are several sign errors in Sundman (1991). You can see from the expression for the Gibbs contribution that a negative magnetic moment will be a problem for the np.log term. A negative Curie temperature is also nonsensical.

sannecottaar: Great, I was quite confused.

bobmyhill:
Ok, I've added a unit test.

bobmyhill:
Is this ready to go now?

tjhei:
Ugh, still need to look at it. Sorry busy time of the year.

ian-r-rose:
The benchmark seems to differ from the reference model by ~1 J/mol.  Is that expected?

burnman/eos/hp.py
```diff
@@ -153,7 +153,12 @@ def gibbs_free_energy(self,pressure,temperature, volume, params):
             else:
                 Gdisord=0.0
 
-        return params['H_0'] + self.__intCpdT(temperature, params) - temperature*(params['S_0'] + self.__intCpoverTdT(temperature, params)) + intVdP + Gdisord
+        if params.has_key('magnetic_moment'):
+            Gmagnetic=self._magnetic_gibbs(pressure, temperature, params)
+        else:
+            Gmagnetic=0.0
+
+        return params['H_0'] + self.__intCpdT(temperature, params) - temperature*(params['S_0'] + self.__intCpoverTdT(temperature, params)) + intVdP + Gdisord + Gmagnetic
```
ian-r-rose: Should this stuff be handled in validate_parameters?

bobmyhill:
Regarding the benchmark differences: Yes - the Sundman dataset includes a term with T^-9, which cannot be modelled by the Holland and Powell formalism.

burnman/eos/hp.py
bobmyhill: Oooh, a philosophical question :)
You mean setting the magnetic params to zero if they don't exist? Would that help much? 

In this case, magnetic_moment is just a cue to compute the magnetic gibbs contribution if necessary (i.e. if the phase is magnetic). If "magnetic_moment" was set to zero in validate_params, then the gibbs contribution would be zero, but running the computations would probably be a lot more expensive than running the param check.

ian-r-rose:
Okay, as long as that is expected.  Is there any documentation to that effect?

burnman/eos/hp.py
ian-r-rose: You give me too much philosophical credit.  :)

I don't mean necessarily setting it to zero, I just mean that it would be more consistent to do this kind of check with the others.  I don't really have a specific suggestion for how to handle it, so don't worry about it too much.  

You are right that we don't want to do unnecessary computations.

tjhei:
PR looks good to me. @ian-r-rose ?

ian-r-rose:
Oof, sorry, forgot about this one, Bob.  Looks good to me.

PR #122, opened 2015-03-02
Subregular solution model

bobmyhill:
The activity-composition models currently in burnman are those used by Holland and Powell. These are by far not the only formulations, and indeed Holland and Powell's asymmetric model is somewhat inflexible, especially  when it comes to fitting first derivatives of the Gibbs free energy (like fO2, for example).

The subregular solution model introduces cubic terms to the quadratic regular solution model. It has been used extensively in the thermodynamics literature. A detailed derivation of useful terms can be found in Helffrich and Wood (1989).

### DIFF
--- a/burnman/solutionmodel.py
+++ b/burnman/solutionmodel.py
@@ -361,24 +361,27 @@ class SubregularSolution (IdealSolution):
         #initialize ideal solution model
         IdealSolution.__init__(self, endmembers )
 
+    def _non_ideal_function(W, molar_fractions):
+        # equation (6') of Helffrich and Wood, 1989
+        n=len(molar_fractions)
+        RTlny=np.zeros(n)
+        for l in range(n):
+            val=0.
+            for i in range(0,n):
+                if i != l:
+                    val+=0.5*molar_fractions[i]*(W[l][i]*(1-molar_fractions[l] + molar_fractions[i] + 2.*molar_fractions[l]*(molar_fractions[l] - molar_fractions[i] - 1)) + W[i][l]*(1.-molar_fractions[l] - molar_fractions[i] - 2.*molar_fractions[l]*(molar_fractions[l] - molar_fractions[i] - 1)))
+                    for j in range(i+1,n):
+                        if j != l:
+                            val+=molar_fractions[i]*molar_fractions[j]*(W[i][j]*(molar_fractions[i] - molar_fractions[j] - 0.5) + W[j][i]*(molar_fractions[j] - molar_fractions[i] - 0.5))
+            RTlny[l] = val
+        return RTlny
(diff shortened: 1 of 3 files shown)


bobmyhill:
We shouldn't merge this until after my solid solution reworking (PR #123) has been merged. It will be quicker to rebase this than the reworking itself.

burnman/solutionmodel.py
```diff
@@ -326,3 +323,81 @@ def __init__( self, endmembers, enthalpy_interaction, volume_interaction = None,
...
+        #initialize ideal solution model
+        IdealSolution.__init__(self, endmembers)
+
+    def _non_ideal_function(self, W, molar_fractions):
+        # equation (6') of Helffrich and Wood, 1989
+        n=len(molar_fractions)
+        RTlny=np.zeros(n)
+        for l in range(n):
+            val=0.
+            for i in range(0,n):
```
tjhei: range(0,n) is equal to range(n), right?

bobmyhill:
@ian-r-rose Thanks for merging #126 !

I think this one should also be ready to roll :)

bobmyhill:
Any thoughts? Just checking in - I don't want to push if everyone's busy :)

burnman/solutionmodel.py
```diff
@@ -326,3 +323,81 @@ def __init__( self, endmembers, enthalpy_interaction, volume_interaction = None,
...
+        if volume_interaction is not None:
+            for i in range(self.n_endmembers):
+                for j in range(i+1, self.n_endmembers):
+                    self.Wv[i][j]=volume_interaction[i][j-i-1][0]
+                    self.Wv[j][i]=volume_interaction[i][j-i-1][1]
+
+        #initialize ideal solution model
+        IdealSolution.__init__(self, endmembers)
+
+    def _non_ideal_function(self, W, molar_fractions):
```
ian-r-rose: Is it possible to write this function in terms of matrix-vector operations?

bobmyhill: Well, this is how it's written in the original paper. I'm not sure it would be more readable in terms of matrix operations, but I can have a look if you think it's worth it?

ian-r-rose: I meant more in terms of performance.  I have gained a reflex against nested loops in interpreted languages, and this gets called every time you need to evaluate the interaction terms.

Of course, now that I look at it, the Ideal model also has nested loops. Maybe it is best to open an issue and deal with it later.

bobmyhill: Agreed. So you're happy to merge this? I'm about to submit a paper which includes it, and would like to add the files as Supplementary Materials...

PR #126, opened 2015-03-06
Solution model restructuring

bobmyhill:
The solution model structure was difficult to read, and some of the parameters used in __init__ were not retrievable after use. This restructuring improves compactness, readability and functionality.

### DIFF
--- a/burnman/minerals/HP_2011_ds62.py
+++ b/burnman/minerals/HP_2011_ds62.py
@@ -26,18 +26,15 @@ N.B. VERY IMPORTANT: The excess entropy term in the regular solution model has t
 
 class garnet(SolidSolution):
     def __init__(self, molar_fractions=None):
-        # Name
         self.name='garnet'
+        self.endmembers = [[py(), '[Mg]3[Al]2Si3O12'], [alm(), '[Fe]3[Al]2Si3O12'], [gr(), '[Ca]3[Al]2Si3O12'], [maj(), '[Mg]3[Mg1/2Si1/2]2Si3O12']]
+        self.type='asymmetric'
+        self.alphas = [1.0, 1.0, 2.7, 1.0]
+        self.excess_enthalpy=[[2.5e3, 29.1e3, 15e3],[10e3,18e3],[48e3]]
+        self.excess_entropy=[[0., 0., 0.],[0., 0.],[0.]]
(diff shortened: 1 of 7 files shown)


ian-r-rose:
Sorry I'm coming a little late to this discussion.  Overall I don't have any strong objections to this, though I don't really see how it is any simpler than how it was before.  What was wrong with requiring a solid solution to specify the solution model in the arguments?  It seems that this way there is a lot of type checking (or 'hasattr' checking) that we didn't have before.

bobmyhill:
Hi Ian,
I see your points, and to some extent I agree. The code base certainly isn't simpler (it added ~30 lines, including those hasattr checks), but I think the user experience is much simplified now - this PR comes from me working with burnman more or less continuously over the last month-or-so. 

There were five problems that I could see:
1) The order of parameters was not intuitive to me. We normally (i.e. in perplex, thermocalc etc.) define excesses in terms of reference values and then temperature and pressure dependence (i.e. excess entropy and then volume), even though the entropy terms were normally zero. I was worried about users (and me) inputting terms in the wrong order, especially as the models got more complicated.
2) Legibility and usability. I've just made a new solid solution model with two more terms, and passing six terms to a solution model in the init struck me as a) ugly, and b) a disaster waiting to happen. Additionally, remembering the class name of the solid solutions was more difficult than just remembering self.type = ideal, symmetric, asymmetric ... and querying and changing type is more elegant as well.
3) Retrieving input params. This was a pain.
4) Fitting was also a pain. If I just wanted to change one number and reinitialise, I needed to pass all of the original arrays to init again. 
5) Repetition. For example, endmembers were defined as a list of lists, then passed manually to init (twice for each class!), then made into an attribute. It seemed sensible to make them into an attribute, and then let the solutionmodel __init__ do the rest. 

I know there are multiple solutions to the above problems, but the one in this PR seemed most elegant to me.

The hasattr calls aren't vital (and I know they're not very pythonic), but they are useful checks for the user. We didn't have any checks before, essentially assuming that the user would make a mistake-free class every time.

bobmyhill:
Is this ok to be merged? I'm needing the other pull request (the margules branch) more and more, and it would be good to have these two finished.

ian-r-rose:
Seems fine to me.  The subregular solution option will be broken until #122 is merged.  But I don't think that is worth fussing about.

PR #130, opened 2015-06-28
Changed slb eos volume output from list to float

bobmyhill:


tjhei:
So this special case is not used in any example or other code right now?

bobmyhill:
It's not a special case, really. I expect burnman to always return a float when I ask it for Cp, S, H, V etc etc.. When the volume query returned a list with a single float a couple of my scripts didn't work (because some of the scipy optimization procedures take a list of floats as input, rather than a list of list of floats). 

I was a little surprised that this didn't show up in any of the tests - perhaps python is more forgiving than I expected.

tjhei:
Well, it is only executed if the brentq can not be applied.

tjhei:
Anyways, thanks for fixing.

bobmyhill:
Ah, I see what you mean. Yes, it is a bit special in that case :)
(I'm dealing a lot with silicate melts (as formulated by de Koker and Stixrude) at the moment, so I'm pushing some of the models to the very limits of their applicable ranges.

Thanks for merging!

PR #131, opened 2015-07-10
Python 3 compatibility

ian-r-rose:
Opening this for comment.  I thought I'd take a crack at making BurnMan compatible with Python 3.  Most of this was done automatically with the 'futurize' package, along with a few updates I had to do by hand.  As far as I can tell, most stuff works with both python 2 and 3 here, though if anybody has the time/inclination, it would be nice to check it on more than one machine.

Currently there is some sort of bug with the install script with python3 that I don't understand, but in-place execution works fine.

### DIFF
--- a/burnman/averaging_schemes.py
+++ b/burnman/averaging_schemes.py
@@ -1,3 +1,5 @@
+from builtins import range
+from builtins import object
 # BurnMan - a lower mantle toolkit
 # Copyright (C) 2012, 2013, Heister, T., Unterborn, C., Rose, I. and Cottaar, S.
 # Released under GPL v2 or later.
@@ -559,7 +561,7 @@ def voigt_average_function(phase_volume, X):
     voigt_reuss_hill classes, takes a list of
     volumes and moduli, returns a modulus.
     """
-    it = range(len(phase_volume))
+    it = list(range(len(phase_volume)))
     V_i = phase_volume
     V_tot = sum(V_i)
     X_voigt = sum(V_i[i]/V_tot * X[i] for i in it)
@@ -573,7 +575,7 @@ def reuss_average_function(phase_volume, X):
     voigt_reuss_hill classes, takes a list of
     volumes and moduli, returns a modulus.
     """
-    it = range(len(phase_volume))
+    it = list(range(len(phase_volume)))
     V_i = phase_volume
     V_tot = sum(V_i)
     if (min(X)<=0.0):
--- a/burnman/__init__.py
+++ b/burnman/__init__.py
@@ -54,34 +54,35 @@ Acknowledgement and Support:
     Bill McDonough, Quentin Williams, Wendy Panero, and Wolfgang Bangerth.
 
 """
+from __future__ import absolute_import
 
-from version import version as __version__
+from .version import version as __version__
 
 # classes for representing rocks and minerals:
-from mineral import Mineral
-from material import Material
-from composite import Composite
-from solutionmodel import SolutionModel
-from solidsolution import SolidSolution
-from mineral_helpers import *
+from .mineral import Mineral
+from .material import Material
+from .composite import Composite
+from .solutionmodel import SolutionModel
+from .solidsolution import SolidSolution
+from .mineral_helpers import *
 
 # high level functions
-from main import *
-from model import Model
+from .main import *
+from .model import Model
 
 # mineral library
-import minerals
+from . import minerals
 
 # central user tools
-import seismic
-import averaging_schemes
-import eos
+from . import seismic
+from . import averaging_schemes
+from . import eos
 
-import processchemistry
-import chemicalpotentials
-from solutionmodel import SolutionModel
-import geotherm
+from . import processchemistry
+from . import chemicalpotentials
+from .solutionmodel import SolutionModel
+from . import geotherm
 
 # miscellaneous
-import tools
-from partitioning import calculate_partition_coefficient,calculate_phase_percents
+from . import tools
+from .partitioning import calculate_partition_coefficient,calculate_phase_percents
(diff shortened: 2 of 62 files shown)


tjhei:
I get the following with Python2. I guess I don't have the future library?
```
$ ./test.sh
*** running test suite...
Traceback (most recent call last):
  File "tests.py", line 2, in <module>
    from util import BurnManTest
  File "/home/heister/Dropbox/work/burnman-git/tests/util.py", line 1, in <module>
    from builtins import zip
ImportError: No module named builtins
ERROR: unittests failed
$ python --version
Python 2.7.6
```
I think you can work around this by doing
```
try:
    from builtins import zip
except ImportError:
    pass
```

tjhei:
for python3 I had to replace 
```
-execfile('$t')
+exec(open('$t').read())
```
in test.sh

Then I got some errors:
```
WARNING EXAMPLE NOT LISTED: example_inv_murakami.py
  benchmark.py ... ok
  chemical_potential_benchmarks.py ... ok
!  debye.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 34
    print "error %e"%np.linalg.norm((old-new)/new)
                   ^
SyntaxError: invalid syntax
!  endmember_benchmarks.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 31
    print 'Benchmarks for', database, 'database with method', fo.params['equation_of_state']
                         ^
SyntaxError: invalid syntax
  solidsolution_benchmarks.py ... ok
checking examples/ ...
  example_averaging.py ... ok
  example_beginner.py ... ok
!  example_build_planet.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 224
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 0-1: truncated \uXXXX escape
  example_chemical_potentials.py ... ok
  example_compare_all_methods.py ... ok
  *** skipping example_compare_enstpyro.py !
  example_composition.py ... ok
  example_fit_data.py ... ok
  example_geotherms.py ... ok
  example_grid.py ... ok
  *** skipping example_inv_murakami.py !
  example_optimize_pv.py ... ok
  *** skipping example_partition_coef.py !
  *** skipping example_premite_isothermal.py !
  example_seismic.py ... ok
  example_solid_solution.py ... ok
  example_spintransition.py ... ok
  example_user_input_material.py ... ok
!  example_woutput.py ... FAIL
Calculations are done for:
Composite:
  0.95 of
    'burnman.minerals.SLB_2011.mg_fe_perovskite'
  0.05 of
    'burnman.minerals.SLB_2011.ferropericlase'
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 73, in <module>
TypeError: 'str' does not support the buffer interface
checking misc/ ...
  paper_averaging.py ... ok
  paper_benchmark.py ... ok
!  paper_fit_data.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 43
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 0-1: truncated \uXXXX escape
!  paper_incorrect_averaging.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 32
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 0-1: truncated \uXXXX escape
!  paper_onefit.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 254
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 0-1: truncated \uXXXX escape
  *** skipping paper_opt_pv_old.py !
!  paper_opt_pv.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 44
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 0-1: truncated \uXXXX escape
!  paper_uncertain.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 50
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 0-1: truncated \uXXXX escape
!  paper_uncertainty.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "<string>", line 46
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 0-1: truncated \uXXXX escape
  table.py ... ok
   done
```

burnman/averaging_schemes.py
```diff
@@ -1,3 +1,5 @@
+from builtins import range
```
tjhei: I would like the header to be above the includes.

ian-r-rose:
Hmm, you are right, I had thought that this would not require an additional
package.  Well, I'll look into it further, see whether the compatibility
breaking for zip and range actually matters anywhere.  Perhaps those
imports are unimportant.

bobmyhill:
If all files are going to be changed, would this be a good time to change the year and authors in the 
headers?

ian-r-rose:
Sure, though I am probably going to take another crack at this at some point. This introduced some dependencies that I did not quite intend.

QuLogic:
You should not _need_ to use `builtin`; it simply makes things consistent between 2 and 3. But it would be better to just `import *` if you want to go down that route. Then people running exclusively on one version won't be surprised by errors in the other version. Relatedly, you should import everything from `__future__` consistently.

ian-r-rose:
Thanks for the advice, Elliott.  Did you add the Python 3 support to Cartopy?  I must say, the various different options for supporting both python 2.7 and 3.x are somewhat confusing. Any other tips on the best tool for doing this with a minimal footprint?

QuLogic:
Yes, that was me. I usually use `futurize` as you did. At this point, it may even be possible to write code that works on 2&3 _without_ any external libraries, but the devil lies in the subtleties and you need to be sure everyone adheres to them.

You can run the stage 1 fixes (`-1`) to make changes that don't really require any dependencies. For stage 2 fixes (i.e., breaking changes), it's simplest to use a library like `six` or `future`. Either one should be sufficient at this point.

Of course, you could take a more radical step and just drop 2.7 support...

ian-r-rose:
Closing in favor of a do-over.

PR #133, opened 2015-09-11
Full subregular

bobmyhill:
This is another solid solution model based on the subregular solution model. The difference is that this model utilises intermediate compounds to define excess properties as a function of pressure and temperature. 

This scheme is extremely flexible, allowing the user to create (for example) models where excess volumes and bulk moduli are anticorrelated. Such schemes are necessary where seismic velocities are calculated for phases that exhibit excess volumes (such as garnets).


bobmyhill:
Opened for comments. Not for merging (yet!)

PR #134, opened 2015-09-11
Added Tref and Pref to HP EoS 

bobmyhill:
Sometimes standard temperature and pressure are not particularly useful. This is particularly true for two situations:

1) The endmember in question is not stable under room temperature and pressure, or one of the key constraints is a univariant reaction that occurs at some elevated P,T

2) The endmember has a particularly large P-T range. At high temperatures, EoSs tend to fail as dV/dT tends towards infinity. This unphysical behaviour can be neatly circumvented by defining the properties of the mineral relative to some elevated Pref.


bobmyhill:
Opened for comments. Not for merging (yet!)

ian-r-rose:
There is already a function in the EquationOfState base class which checks the reference temperature, and is used in the slb and mgd EoSs. Comments on that way of doing it versus what you have done here?

Otherwise looks very reasonable to me.

bobmyhill:
I'd forgotten that Nick added those functions...

I have no particularly strong views on this, although having __init__ create a T_0 key for the params dictionary must be faster than all of those function calls and if statements. We should be consistent, whichever way we choose. 
(I also notice that T_ref in the reference_temperature function is 300 K ... I assume this is the SLB Tref. The HP one is 298.15 K...)

Any strong opinions? I'm happy to make the necessary changes ... it won't take long.

ian-r-rose:
Yeah, I think you are probably correct on both counts. Having that in the base class unnecessarily hard-codes 300 K, so it is probably better for each EoS to do it itself. And that is a lot of unnecessary checking the params object. So your way is probably better.

Also, if I recall correctly, 300 K is indeed the reference temperature for SLB, though I can't imagine it matters except for benchmarking purposes.

bobmyhill:
Ok, these should be fixed now. Could someone check that I haven't made any stupid mistakes? The test output is the same as it is for master (which is slightly broken for me, but maybe this is a machine problem?).

burnman/eos/cork.py
```diff
@@ -131,6 +131,11 @@ def validate_parameters(self, params):
         Check for existence and validity of the parameters
         """
 
+        if 'T_0' not in params:
+            params['T_0'] = 298.15
+        if 'P_0' not in params:
+            params['P_0'] = 0.
```
ian-r-rose: It looks like you are changing the default reference pressure from 10^5 Pa to 0 Pa.  Is this intended?

burnman/eos/mie_grueneisen_debye.py
```diff
@@ -174,6 +174,10 @@ def validate_parameters(self, params):
         """
         Check for existence and validity of the parameters
         """
+        if 'T_0' not in params:
+            params['T_0'] = 300.
+        if 'P_0' not in params:
+            params['P_0'] = 0.
```
ian-r-rose: Is this used anywhere in this EoS? My recollection is that all the pressure stuff is turfed off to the isothermal Birch-Murnaghan.

burnman/eos/cork.py
bobmyhill: P_0 is never used in the original expressions, so introducing a standard state pressure of 1.e5 would produce a (small) difference between burnman output and the original results. I favored this alternative...

burnman/eos/mie_grueneisen_debye.py
bobmyhill: Yes, that's right. But as we don't currently run bm.validate_parameters() if we're using the SLB/MGD eos, I think this is the best place to initialise the parameter. Happy to hear alternatives though, if you have a better idea...

examples/example_thermodynamic_beginner.py
```diff
@@ -0,0 +1,184 @@
+# BurnMan - a lower mantle toolkit
```
ian-r-rose: Can we break this new example into a separate PR? I think it needs some commenting to tell what is going on.

burnman/eos/mie_grueneisen_debye.py
ian-r-rose: Hm, I wonder if that is an oversight on my part. Should we call bm.validate_parameters() inside of mgd.validate_parameters()?

examples/example_thermodynamic_beginner.py
bobmyhill: Sure, no problemo.

burnman/eos/modified_tait.py
```diff
@@ -27,7 +25,7 @@ def modified_tait(x, params):
     EQ 2 from Holland and Powell, 2011
     """
     a, b, c = tait_constants(params)
-    return (np.power((x + a - 1.) / a, -1./c) - 1.)/b
+    return (np.power((x + a - 1.) / a, -1./c) - 1.)/b + params['P_0']
```
ian-r-rose: Same here: this is the more correct form of the EoS?

burnman/eos/mie_grueneisen_debye.py
bobmyhill: Hmmm, yes, probably. Good idea!

burnman/eos/hp.py
```diff
@@ -326,6 +324,11 @@ def validate_parameters(self, params):
         if params['Gprime_0'] is not float('nan') and (params['Gprime_0'] < -5. or params['Gprime_0'] > 10.):
             warnings.warn( 'Unusual value for Gprime_0', stacklevel=2 )
 
+        if params['T_0'] < 0.:
+            warnings.warn( 'Unusual value for T_0', stacklevel=2 )
+        if params['P_0'] < 0.:
+            warnings.warn( 'Unusual value for P_0', stacklevel=2 )
```
ian-r-rose: I like these checks. Should we put them in the other relevant EoSs as well?

burnman/eos/modified_tait.py
bobmyhill: Yes; at [Pref, Tref], x = V/V0 = 1, so to return Pref, we need to add it.

burnman/eos/hp.py
bobmyhill: Yep, now added

burnman/eos/birch_murnaghan.py
```diff
@@ -139,28 +139,29 @@ def validate_parameters(self, params):
...
-        #now check that the values are reasonable.  I mostly just
-        #made up these values from experience, and we are only 
-        #raising a warning.  Better way to do this? [IR]
+        #now check that the values are reasonable.
         if params['V_0'] < 1.e-7 or params['V_0'] > 1.e-3:
             warnings.warn( 'Unusual value for V_0', stacklevel=2 )
         if params['K_0'] < 1.e9 or params['K_0'] > 1.e13:
             warnings.warn( 'Unusual value for K_0' , stacklevel=2)
-        if params['Kprime_0'] < -5. or params['Kprime_0'] > 10.:
+        if params['Kprime_0'] < 0. or params['Kprime_0'] > 10.:
```
ian-r-rose: I remember sitting down thinking about whether there was any reason for K-prime to be thermodynamically required to be positive, and could not come up with one.  I mean, eventually it has to be positive, on enough compression, but does this need to be true over the entire range of compressions?

bobmyhill: Hmmm, it's a good question. I don't think there's any requirement that it must be positive, but a negative value would probably indicate a collapse in structure, and I doubt we'll have many users interested in such weird behaviour.

Actually, I changed this to zero because there was a test that checked for a negative K' in SLB/MGD. I think having this warning is useful, and the user will know when to ignore it if they have a strange material.

ian-r-rose:
As a side note, I cannot properly run the tests without #136.  If you rebase I can verifty that everything works on my machine.

burnman/eos/birch_murnaghan.py
ian-r-rose: Sure, in practice it is not too likely to come up, and if it does then a warning seems sensible.  I think it is fine.

bobmyhill:
Ok, rebased... fingers crossed!

ian-r-rose:
Okay, so some tests fail.
`example_chemical_potentials.py`  has a small (1.e-5) difference in log(fO2)
`HHPH_2013.mwd` runs afoul of parameter validation for `V_0`, for some reason.
`paper_onefit.py` outputs a table of parameters, so that is different (I can take a crack at fixing that one)

ian-r-rose:
Can you cherry-pick this commit from my repo?
ian-r-rose@5322f5a35b0ca78b75fb73f5502c4d191376f494

ian-r-rose:
Ah, I see. The warning is actually with `HHPH_2013.msnal`. The modified Tait EoS has a stricter check on reference volume than HP, so now it is catching this mineral.  Is this reference volume correct?  If so, should we relax the validation in MT?

bobmyhill:
Ok, everything is good for me, apart from the errors in paper_onefit.py, which are switched for me, i.e.:

< errors: 6.04361957868e-06 1.96879806264e-07 5.05950473651e-06 7.87235035907e-07
---
> errors: 6.04361785322e-06 7.87264629724e-07 5.05947174541e-06 1.96881478083e-07

Any ideas?

tjhei:
> Any ideas?

I guess you don't have numdiff installed? This is also broken on master, probably we switched the order of the errors at some point. With numdiff the test doesn't fail though (because both are 10^-7).

ian-r-rose:
Everything passes for me. I suppose the output for `paper_onefit.py` needs to be updated, but I don't think it's necessary for this PR.

burnman/eos/hp.py
```diff
@@ -247,35 +234,36 @@ def __thermal_pressure(self,T,params):
         # constant over a wide range of compressions.
 
         # Note that the xi function in HP2011 is just the Einstein heat capacity
-        # divided by 3nR.  I don't know why they don't use that, but anyhow...
+        # divided by 3nR. This function is *not* used to calculate the
```
ian-r-rose: Nice elaboration :)

bobmyhill:
I've just fixed paper_onefit.py.out, so we should be good to go now. Thanks Ian!

PR #136, opened 2015-09-16
When running tests, don't call savefig.

ian-r-rose:
The do-nothing `Template` backend appears to be bugged (as far as I can tell) for at least a few versions of matplotlib when using TeX rendering. Since we don't care about the output when running tests, we can just not call savefig().

### DIFF
--- a/misc/paper_averaging.py
+++ b/misc/paper_averaging.py
@@ -150,5 +150,6 @@ if __name__ == "__main__":
 
     plt.xlabel('Pressure (GPa)')
     plt.ylabel('Shear velocity $V_s$ (km/s)')
-    plt.savefig("example_averaging.pdf",bbox_inches='tight')
+    if "RUNNING_TESTS" not in globals():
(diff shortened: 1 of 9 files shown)


tjhei:
The reason I implemented this back then was that I didn't want to modify the examples with test related information. Too bad this doesn't work anymore.

What happens if you run the old code (it works on my machine)?

ian-r-rose:
The good news is that the examples are all unmodified, the only changes are to misc/paper_*

Testing the master branch on my machine (and possibly Sanne's?) results in an AttributeError which is due to some uninitialized variables in the `Template` backend. It is a simple enough fix in matplotlib, but this seems like a reasonable fix (if a bit hacky) at the moment.

tjhei:
Oh! How come the examples work but the misc/ files don't?

ian-r-rose:
The error appears with plt.savefig(), but not with plt.show() for some reason.  It also only appears when TeX rendering is used. I don't really think it is worth the time to investigate too deeply :)

PR #138, opened 2015-09-28
Updated Composite

bobmyhill:
The proposed Assemblage class can instead be included in the existing Composite class iff the order of phases and fractions are switched in Composite and unroll. The old way of doing things (with phase fractions being mandatory but phases being optional) seems less logical to me.

I realise that this will break all user defined composites, and may therefore meet with some resistance. We can include a two or three line try/except to catch and fix this problem, or just stick with the Exception message that I've added.

If this change meets with approval, then I can work on implementing all the other parts of Assemblage within the Composite class.

### DIFF
--- a/burnman/composite.py
+++ b/burnman/composite.py
@@ -5,9 +5,11 @@
 import numpy as np
 import warnings
 
-from burnman import averaging_schemes
 from burnman import Material
 from burnman import Mineral
+
+from burnman import averaging_schemes
+from burnman import chemicalpotentials
 
 def check_pairs(phases, fractions):
         if len(fractions) < 1:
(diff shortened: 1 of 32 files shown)


tjhei:
> with phase fractions being mandatory but phases being optional

I think you misunderstood the code (not documented very well I have to admit): ```Composite.__init__(self, phases, fractions=None)``` allows you to either do

```Composite([0.2, 0.8], [m1, m2])``` or ```Composite([(m1,0.2), (m2,0.8)])```

for backwards compatibility. You always had to define phases and fractions in a composite.

What is the reason to switch the order of the return values of ```unroll()```?

What are you trying to achieve with ```Assemblage``` and what is the difference to ```Composite```? I see ```Composite``` as a simple container with fixed fractions. If you need to do something more complicated, just derive from ```Material```.

bobmyhill:
I think I understood the code ok - there are examples of both of these formulations in the examples.

- I can understand wanting to keep things backwards compatible, but the [fractions], [phases] ordering in unroll is not consistent with the rest of the code (the old mineral helpers, for example). It makes much more sense to me to declare a bunch of phases and then their fractions, rather than a set of fractions and then phases. I also want to add a fraction_type to the Composite class (i.e. molar, volume ...), which would make the new ordering much more reasonable. 
- If we want to change this (and any other bits of code) I would suggest we do it soon, before more people use it. At the moment, I think there are perhaps 1 or 2 active users in addition to us, but I suspect we will get more people when the papers from Sanne and myself are published.
- The changes suggested here allow one to define a composite of minerals without fractions - sometimes fractions are unimportant (for crystal chemistry, for example). I outlined this in the original Assemblage Class pull request. 
- This pull request is an alternative to the Assemblage Class.

bobmyhill:
Of course, it's true that if fractions are not important, one could just put in any values to the composite class. I just find this relatively minor reworking much more elegant.

sannecottaar:
I agree with Bob. We should agree on one formulation and change all the examples. The current release won't be backward compatible as it is. I think this is our last chance to make good choices, and defining a composite with only the phases seems reasonable. 
Bob, where you planning some sort of set_fraction function too? This could make some examples more elegant where we loop through fractions, but now have to redefine a composite every time.

bobmyhill:
Hi Sanne,

Yes, I've added set_fractions and some examples of use to example_composite.py, I need to finish adding the moduli and thermodynamic functions, but the core functionality is here now.

sannecottaar:
I think we should merge this. Any other opinions?

bobmyhill:
Not yet! The seismic velocity stuff still needs a bit of work... and I think the tests may be broken.

sannecottaar:
I was thinking of updating composite after a merge, so one can call moduli,
seismic velocities etc. for the composite (either the composite as a whole,
or the separate phases in the composite). And also include an 'evaluate'
function to do a list of them. Quite similar to how seismic models are
structured. This would remove a bunch of functions from main.py.
Or is that similar to what you are planning?

ian-r-rose:
I think this is an okay change, with all the caveats discussed above (especially about this being breaking). If we do this, we should make it very clear which functions can be called without the `fractions` being set.

Bob, the internal functions `molar_to_mass_fraction` etc. look more generally useful (I have wanted those types of things more than once). What would you think about exposing them as utility functions?

I like Sanne's proposal about averaging moduli, velocities, etc being within the composite class.

tjhei:
> I like Sanne's proposal about averaging moduli, velocities, etc being within the composite class.

I don't quite understand why you want to put stuff into Composite. Composite is a ```Material``` so functionality should go into the base class if possible. Then computing things magically works for Mineral, Composite, and anything else we might have in the future.

I still like the idea of keeping Composite as simple as possible ("give me 80% pv and 20% fp") and introduce new classes derived from Material f you need to do something more fancy.

ian-r-rose:
Ah, you are absolutely correct, Timo. Since `Material` defines `unroll()`, and that is all that is required for averaging schemes, that is where it should go.

bobmyhill:
Ok, I don't have time to rearrange set_state now, but I agree that that's how it should be done (from Material on down). Sorry Timo, I didn't understand you when you first suggested this back in ?January. I've uploaded a copy of this branch to upstream if someone wants to play.

I don't think Composite is becoming too fancy - we're just allowing the fractions to change. 

I wanted to add the chemical potential functionality because it will make thermodynamics coding much more elegant, but I realise that it wouldn't be much work on the user's part to make their own little list of phases and call the function I already made.

sannecottaar:
I'm playing around with allowing every material property to be computed for every material. Nothing will be computed in set_state, so this will save unused computations, or attempting computations that won't work for given minerals. (I hope nobody else has started on this. )

I have a more general question on the philosophy of set_state. I currently want to implement things in the following way: 
 - for a given material Mat, one can set the state to a pressure and temperature (Mat.set_state(P,T)) and then request all kinds of properties (Mat.V(), Mat.K_T()). 
 - for a given material Mat, one can ask for properties over a range of pressures and temperatures (Mat.V([P],[T]), Mat.K_T([P],[T]). In this case arrays are returned, and the 'state' won't be set to the last value of P and T (like currently happens when you loop over values in main.py). 

Alternatively we could allow for 'arrays of states' in set_state? 
Or not allow arrays at all and people have to loop in their own code?

Do people have a preferred or different option?

bobmyhill:
I had started on this (my fork, branch "attribute_efficiency"), but I don't mind you redoing it. There are a couple of ideas in the branch:
- Changing the function variables to (self, mineral)
- Wiping properties on each set_state
- Having a try: except: for each property.

I remember that ?Timo/?Ian mentioned a great way of creating attributes on the fly, which was more elegant than my attempt in the branch "attribute_efficiency". The idea is to add "@property" before a function (e.g. def H()) to make it a property of the class (see http://www.python-course.eu/python3_properties.php). That way, you don't need Mat.V() - we can stick with Mat.V and not break everyone's code.

I'd say that having an array of states risks making things very messy. For example, the way you describe things for volume calculations makes set_state redundant and means that we'd need two different V functions. I'd far rather the user loops through P-T points themselves, or we build some higher level functions to run arrays. But I'm open to the alternative if you have a good argument for it.

sannecottaar:
I'll have a look at this. I started with your update_composite branch.

Having it as a property would be a lot more elegant!

I was planning to have both option in one function (either P and T are
given, or the set_state is used). But you are right about messy. Just now
thinking it is easier to build  an evaluate function that can handle a list
of properties as well as arrays of P, T  (similar as in seismic models).

bobmyhill:
Yes, I think that would be a great idea. Something like this perhaps?:
def evaluate(mineral/rock/squirrel, pressures, temperatures, ['V', 'K_T']):

Somewhere there will have to be a mapping from the strings to properties, but this should still be significantly more elegant than messing around with the functions in mineral.py.

PR #140, opened 2015-10-08
fix optimize perovskite example

sannecottaar:
Just noticed that optimize_pv from the paper is still using the old solid solutions. While the example of optimize_pv (in the examples directory) just has a really poor choice of composition. I've tried to update that example here a bit. Please check so I can add the code to our slides.

### DIFF
--- a/examples/example_optimize_pv.py
+++ b/examples/example_optimize_pv.py
@@ -36,10 +36,6 @@ from burnman import minerals
 
 if __name__ == "__main__":
 
-
-### input variables ###
-    #######################
-
     # Define reference model and depth to evaluate
     seismic_model = burnman.seismic.PREM()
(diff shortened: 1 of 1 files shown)


ian-r-rose:
Seems reasonable. Two questions:
* does it make sense to plot the redundant K and G?
* what is the effect of normalizing the errors here? The original benefit of just having wave velocities was that the errors had the same units.

sannecottaar:
Fitting all the different parameters leads to a different result in optimal perovskite fraction. E.g. one can fit V_s, but not necessarily be fitting rho and G and the same time. 
The normalizing is for the purpose of plotting (I won't be showing that this evening).

ian-r-rose:
Okay, sounds good.

PR #142, opened 2015-10-12
remove absolute numdiff tolerance

tjhei:
This caused the test misc/paper_onefit to pass where we swapped values a
long time ago because both numbers where ~1e-7.


ian-r-rose:
Oops, just merged Bob's PR which had the same update to some test results. Can you regenerate them?

tjhei:
easy. :-)

ian-r-rose:
Lovely.

PR #144, opened 2015-10-28
Added activity functions to solid solution, updated example document

bobmyhill:
Quite often, we want to use the activity of a particular endmember in a solid solution, rather than the excess gibbs free energy. This commit adds two attributes to solidsolution, activities and activity_coefficients, which output the overall activity and the proportion of this value attributable to nonideality.

### DIFF
--- a/burnman/solidsolution.py
+++ b/burnman/solidsolution.py
@@ -126,23 +126,16 @@ class SolidSolution(Mineral):
     @material_property
     def activities(self):
         """
-        Returns a list of endmember activities
+        Returns a list of endmember activities [unitless]
         """
-        if self.temperature > 1.e-10:
-            return np.exp(self.excess_partial_gibbs/(constants.gas_constant*self.temperature))
-        else:
-            return None
+        return self.solution_model.activities( self.pressure, self.temperature, self.molar_fractions)
 
     @material_property
     def activity_coefficients(self):
         """
-        Returns a list of endmember activity coefficients (the non-ideal part of the activities
+        Returns a list of endmember activity coefficients (gamma = activity / ideal activity) [unitless]
         """
-        if self.temperature > 1.e-10:
-            return np.exp(self.excess_partial_gibbs/(constants.gas_constant*self.temperature)
-                          - IdealSolution._log_ideal_activities(self.solution_model, self.molar_fractions))
-        else:
-            return None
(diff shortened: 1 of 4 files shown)


sannecottaar:
Looks good to me, as far as I understand this. It is spitting out a runtime warning when temperature is 0 for the benchmark_solidsolution.py, so the output should be fixed to pass that test.

bobmyhill:
Fixed!

bobmyhill:
Ok, should be good to go...

burnman/solidsolution.py
```diff
@@ -122,9 +121,30 @@ def set_state(self, pressure, temperature):
...
+        """
+        if self.temperature > 1.e-10:
+            return np.exp(self.excess_partial_gibbs/(constants.gas_constant*self.temperature))
+        else:
+            return None
 
+    @material_property
+    def activity_coefficients(self):
+        """
+        Returns a list of endmember activity coefficients (the non-ideal part of the activities
```
tjhei: closing parens?

burnman/solidsolution.py
```diff
@@ -122,9 +121,30 @@ def set_state(self, pressure, temperature):
         Mineral.set_state(self, pressure, temperature)
         for i in range(self.n_endmembers):
             self.endmembers[i][0].set_state(pressure, temperature)
+            
 
+    @material_property
+    def activities(self):
+        """
+        Returns a list of endmember activities
```
tjhei: units?

tjhei:
it would also be great to have a test that somehow covers numeric values of these functions.

bobmyhill:
Ok, I've changed the implementation to be a bit more efficient (there's no np.exp(np.log()) to calculate activities now). I've also added some tests and added internal and helmholtz energies to the solid solution calculations.

tjhei:
I still have no idea what these activities are, but I am not a mineral physicist. :-)

Looks good to me.

bobmyhill:
I suspected that this was the reason that it wasn't merged yet :) There's a good introduction here:
http://serc.carleton.edu/research_education/equilibria/activitymodels.html

Don't worry, I'll add some details to the documentation in a separate PR. Activities are very useful.

bobmyhill:
@sannecottaar, are you also happy?

sannecottaar:
Yes. I'm just going to go ahead and merge this.

PR #147, opened 2015-11-16
Correct landau V, changed SLB references to T_0

bobmyhill:
This patch fixes the volume Landau contribution, removes the double counting of the enthalpy of disorder (oops), and completes the change of references to T_0 in SLB.

There are still changes to make w.r.t. Landau (w.r.t. K_T, K_S), but these can wait until our redesign of excess contributions in December.

### DIFF
--- a/burnman/endmemberdisorder.py
+++ b/burnman/endmemberdisorder.py
@@ -75,10 +75,22 @@ def enthalpy_disorder_Landau(P, T, params):
     from Holland and Powell, 1996, corrected using
     landaunote.pdf on Tim Holland's esc web page
     """
-    # N.B. Assumes Vt==1, see above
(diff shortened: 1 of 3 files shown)


ian-r-rose:
Looks good to me. Are the functions at the end of `endmemberdisorder.py` meant to be user-facing? In either case, can you annotate where they come from? I am assuming HP1996?

bobmyhill:
Hi Ian!

They can be user facing, although I don't think they will be used much. I've now added the annotations.

This PR is very much a temporary fix before we rework/optimize the code next month. I think these optimizations should be a primary aim for us - I've had a couple of comments about how slowly the code runs. For now, this fixes a couple of moderate issues.

ian-r-rose:
Cool, thanks Bob.

I agree that optimizations should be a priority.

PR #151, opened 2015-12-10
Speed up slb

sannecottaar:
This speeds up solving for the volume in slb (which is quite a bottleneck) and the debye function (which just gets called a lot). Not as significant of an improvement to adding properties.

### DIFF
--- a/burnman/eos/slb.py
+++ b/burnman/eos/slb.py
@@ -64,33 +64,46 @@ class SLBBase(eos.EquationOfState):
         P_th = gr * debye.thermal_energy(T,Debye_T, params['n'])/V
         return P_th
 
+    def _volume(self,x,pressure,temperature,V_0,T_0,Debye_0,n,a1_ii,a2_iikk,b_iikk,b_iikkmm):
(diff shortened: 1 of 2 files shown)


burnman/eos/slb.py
```diff
@@ -68,33 +68,46 @@ def __thermal_pressure(self,T,V, params):
...
         # we need to have a sign change in [a,b] to find a zero. Let us start with a
         # conservative guess:
         a = 0.6*params['V_0']
         b = 1.2*params['V_0']
-
+        args=pressure,temperature,V_0,T_0,Debye_0,n,a1_ii,a2_iikk,b_iikk,b_iikkmm
         # if we have a sign change, we are done:
-        if func(a)*func(b)<0:
-            return opt.brentq(func, a, b)
+        vol_a =self.__volume(a,pressure,temperature,V_0,T_0,Debye_0,n,a1_ii,a2_iikk,b_iikk,b_iikkmm)
```
tjhei: use __volume(a,args) here?

PR #152, opened 2015-12-10
Remove double function

sannecottaar:
This removed the double function for pressure in mgd (which were similar)  and a number of other errors highlighted by pylint.


tjhei:
:+1:

PR #162, opened 2015-12-11
New tools

bobmyhill:
- Added generic array_to_file function

### DIFF
--- a/burnman/tools.py
+++ b/burnman/tools.py
@@ -90,6 +90,43 @@ def array_from_file(filename):
     data = np.array(zip(*data))
     return data
 
+def array_to_file(labels, array, filename):
+    """
+    Generic function to read a list of lists into a file.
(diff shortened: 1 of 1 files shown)


burnman/tools.py
```diff
@@ -90,6 +90,43 @@ def array_from_file(filename):
     data = np.array(zip(*data))
     return data
 
+def array_to_file(labels, array, filename):
+    """
+    Generic function to read a list of lists into a file.
```
tjhei: "read"? You mean like, write?

burnman/tools.py
```diff
@@ -90,6 +90,43 @@ def array_from_file(filename):
...
+    Includes a one-line label header initiated with a 
+    % character.
```
tjhei: headers are typically encoded with "#", any specific reason to go with %?

PR #167, opened 2015-12-11
Test that it is possible to create all the minerals in the mineral li…

ian-r-rose:
…brary and call set_state() on them.


tjhei:
looks good, merge yourself if you are happy with it

PR #169, opened 2015-12-12
fix and test seismic

tjhei:
- fix fast/slow density
- improve error message if density is not defined
- fix comments
- test!

### DIFF
--- a/burnman/seismic.py
+++ b/burnman/seismic.py
@@ -292,7 +292,8 @@ class SeismicTable(Seismic1DModel):
 
 
     def density(self, depth):
-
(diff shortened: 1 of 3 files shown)


burnman/seismic.py
```diff
@@ -292,7 +292,8 @@ def QG(self, depth):
 
 
     def density(self, depth):
-
+        if len(self.table_density)==0:
+            raise ValueError("Density has not been defined for this seismic object")
```
ian-r-rose: Seismic model instead of seismic object

tjhei:
fixed!

PR #197, opened 2015-12-13
Fix ipython

sannecottaar:


ian-r-rose:
The documentation for the equations of state is incorrect. bm2, slb2, and mgd2 are all third order equations of state with second order expansions for shear modulus. Not your mistake, but you touched it last!

tjhei:
(do not merge yet)

sannecottaar:
I think this can be merged...

sannecottaar:
Although maybe not, because currently this works for python3, but not python2.7...

tjhei:
Urgh, ipython stuff is a mess. So the test script I made is not working correctly and not detecting that the conversion fails. Another issues is python2 vs python3 as you said. Should we go with python3 only?

tjhei:
Can you take over https://github.com/tjhei/burnman/commit/66e36aac911d1346c6710c5c4925e391c33b356b please? That only runs the ipython tests under python3 and only if found.

tjhei:
Oops. Better use https://github.com/tjhei/burnman/commit/07f74b346f4ef43df989abb3c4b6335e0c552a35

tjhei:
and the ipython version is on the tester is too old to convert the files. :-(

sannecottaar:
We should maybe just not test these. Shall we like we discussed at AGU make a folder for 'contributes scripts', maybe with subfolder for published work and educational material? Than we can put the ipython notebooks in there (and in general not test, or provide upkeep on this directory. Hopefully we keep things backward compatible from now on).

sannecottaar:
I remove the ipython folder from the tests, since there are too many issues with versions etc. I propose moving this to the educational folder later. 
However there are still a number of actual fixes and updates in the notebook files hidden in this pull request that should probably be merged.

bobmyhill:
Is this good to go now?

sannecottaar:
Yes, it's not being tested though, but the corrections should go through
for the new version.

PR #218, opened 2015-12-18
fix murakami example

tjhei:


tjhei:
is the setup of the molar fractions correct like this?

bobmyhill:
Hi Timo,
I suspect not. The bounds on b and c are set to be low (0.0 - 0.2), which suggests that it is Fe proportion that corresponds to the parameters. The order of molar fractions is MgO, FeO and MgSiO3, FeSiO3, AlAlO3 for periclase and perovskite respectively.

Other things:
- It would be good to rename the parameters to something less obscure (X_FeO and X_FeSiO3, for example).
- It would also be good to change the name of perovskite to mg_fe_al_perovskite.

bobmyhill:
@tjhei What's the status of this PR? It wouldn't take much to make these changes, and it would be nice to get all of these little things merged soon so that we can start on the big stuff.

tjhei:
updated and fixed. This is the output I get after a few minutes:

example_inv_murakami

the best fit is this:
example_inv_murakami_2

PR #222, opened 2015-12-28
tune performance

tjhei:
- jit some functions
- add misc/performance.py with several checks

This patch increases performance from:
         4331518 function calls (4024502 primitive calls) in 6.741
seconds
to
         3574686 function calls (3263157 primitive calls) in 4.139
seconds

### DIFF
--- a/burnman/eos/slb.py
+++ b/burnman/eos/slb.py
@@ -33,20 +33,19 @@ def _grueneisen_parameter_fast(V_0, volume, gruen_0, q_0):
     nu_o_nu0_sq = 1.+ a1_ii*f + (1./2.)*a2_iikk * f*f # EQ 41
     return 1./6./nu_o_nu0_sq * (2.*f+1.) * ( a1_ii + a2_iikk*f )
 
-@jit
-def _delta_pressure_fast1(V_0, a1_ii, a2_iikk, x, Debye_0):
-    f = 0.5 * (pow(V_0 / x, 2. / 3.) - 1.)
-    debye_temperature = Debye_0 * np.sqrt(1. + a1_ii * f + 1. / 2. * a2_iikk * f * f)
-    return debye_temperature, f
-
 
 @jit
-def _delta_pressure_fast2(E_th, E_th_ref, a1_ii, a2_iikk, b_iikk, b_iikkmm, f, pressure, x):
-    nu_o_nu0_sq = 1. + a1_ii * f + (1. / 2.) * a2_iikk * f * f  # EQ 41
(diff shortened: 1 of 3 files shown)


misc/performance.py
```diff
@@ -0,0 +1,75 @@
...
+        x = 0
+        for i in range(0,10000):
+            x += a.grueneisen_parameter(1e9,1e4,1.1,m.params)
+    return x
+
+import timeit
+
+if False:
+    r=timeit.timeit('__main__.test_grueneisen()', setup="import __main__", number=3)
+    print("test_grueneisen", r)
```
sannecottaar: Timo, why do we skip this? And keep it in?

tjhei: I used this for testing if my changes had a positive impact. I agree it is not great to have it in here as is. I will update it.

tjhei:
okay, updated.

sannecottaar:
Looks good to me.

burnman/eos/slb.py
```diff
@@ -7,12 +7,46 @@
...
+    """global function with plain parameters so jit will work"""
+    x = V_0 / volume
+    f = 1./2. * (pow(x, 2./3.) - 1.)
+    a1_ii = 6. * gruen_0 # EQ 47
+    a2_iikk = -12.*gruen_0 + 36.*gruen_0*gruen_0 - 18.*q_0*gruen_0 # EQ 47
+    nu_o_nu0_sq = 1.+ a1_ii*f + (1./2.)*a2_iikk * f*f # EQ 41
+    return 1./6./nu_o_nu0_sq * (2.*f+1.) * ( a1_ii + a2_iikk*f )
+
+@jit
+def _delta_pressure_fast1(V_0, a1_ii, a2_iikk, x, Debye_0):
```
ian-r-rose: This function seems mis-named. Perhaps `_debye_temperature_fast`?

burnman/eos/slb.py
```diff
@@ -70,18 +104,13 @@ def _thermal_pressure(self,T,V, params):
         gr = self.grueneisen_parameter(0., T, V, params) # P not important
         P_th = gr * debye.thermal_energy(T,Debye_T, params['n'])/V
         return P_th
-    
+
     def _delta_pressure(self,x,pressure,temperature,V_0,T_0,Debye_0,n,a1_ii,a2_iikk,b_iikk,b_iikkmm):
```
ian-r-rose: Can you describe why it is a performance improvement to split up the function this way and call `_delta_pressure_fast1` and `delta_pressure_fast2`? Can one jitted function not call another jitted function?

tjhei: Because I messed up when I tried that. Fixed now.

ian-r-rose:
Looks good to me, thanks!

PR #225, opened 2016-01-05
remove lost function for attenuation correction

sannecottaar:
This function never gets used, and only loops over a different (equally unused) function anyway.


tjhei:
what happened to attenuation correction? does this live somewhere else now?

sannecottaar:
There is a function in seismic. This one only loops over a list. Neither of
them are used. I think I will improve this at some point, but this function
is in the wrong location, useless and inflexible at the moment.

PR #226, opened 2016-01-05
Prettifying code

sannecottaar:
Autopep8 cleans up the code. Does anyone know of a way to commit things without authorship? As this is a bit ridiculous. I know people are watching the line modifications stats ;).  
I think this is nice to do before release (which I think we can do now, or are there still big issues?)


tjhei:
> anyone know of a way to commit things without authorship? 

Yes.

But we should wait until all other PRs are merged or closed, because you end up with a ton of conflicts otherwise.

tjhei:
see #244

PR #227, opened 2016-01-11
Fix number of volume calls

sannecottaar:
I noticed the code had slowed down (while working on updating the inversion branch), because a couple more volume class snuck in with the modifications to properties. I'm not sure if this is the most elegant way to solve it, so I'm open to suggestions.


sannecottaar:
This pull request should also be looked at, since it will speed up the code.

bobmyhill:
I remember we had several functions which were called with 0. as the pressure or volume input (depending on whether the EoS was V, T or P, T). 

I still don't really like calling functions with dependent variables, but as this now seems to be the norm in the code, I guess I'm ok with it.

If you want to make this change, all the EoSs should be changed consistently. I think the tests will cover all the potential pitfalls...

tjhei:
Looks good to me!

PR #228, opened 2016-01-14
Inversion

sannecottaar:
Updating the branch and the basic inversion setups (how solid solutions are defined, the new order in Composite and the new evaluate functions).


bobmyhill:
@sannecottaar The merges (rather than rebases) have made this old branch unreviewable. Is it ok if we close this PR? More generally, is this something that can be reworked? Khan and Connolly published an uncertainty analysis on the SLB dataset using MC methods recently. 

I'd argue that any analysis of uncertainties using the SLB database requires the full covariance matrix from Lars and Carolina; do you think this is something they might let you/us use?

bobmyhill:
I'm closing this PR for now.

PR #229, opened 2016-02-09
Contributions folder

sannecottaar:
Here is one of the things that we discussed during lunch at AGU: having a folder with published and educational contributions (which we don't generally test or are responsible for, but which we can hopefully keep backward compatible for now on). 
For the moment I put the tutorial in there, and the scripts related to our publication. I originally wanted to propose removing the scrips from the other locations, but unfortunately our paper states that the scripts are in the misc directory, so maybe we should keep them there. I'm also fine taking it out of this folder again. 
Cayman, your planetary builder is directly linked to your paper, right? It should probably go in here.


tjhei:
- are the tutorials now duplicated in /tutorial/ ? 
- is there a reason you didn't move the paper scripts from /misc/ to the new folders?
- do we need to test the new scripts? I think yes!
- if we accept contributions we need to think about licensing issues or clearly state that the contributions folder is not part of BurnMan released under GPL (and exclude it from releases). The current PR is fine of course, because it is our data.

sannecottaar:
> are the tutorials now duplicated in /tutorial/ ?
Yes, but happy to remove them is this setup is approved. And then also change the directories being tested. 

> is there a reason you didn't move the paper scripts from /misc/ to the new folders?
Mainly because they are referenced in the paper as being in misc? I could maybe just put symbolic links in the misc folder?

bobmyhill:
In the readme, I'd change "main authors" to "development team (see Readme.md for a list of developers)". Also, some more guidance for new contributors will probably save time in the long run. For example, each contribution should include a readme.

I agree that contributions should be tested. Perhaps there should be a master script in each contributions directory that runs those *.py files which should be tested, and a test_contributions.sh in the main contributions directory that loops through each contribution, analogous to the test.sh in the main directory.

If previous contributions no longer provide the same result, or if they are broken, they should be removed from the relevant master script, and a note should be added to the readme explaining the reason for the change in result.

Otherwise, this looks good!

sannecottaar:
I cleaned it out a bit more. Moved all the files (this means the scripts won't be in misc anymore, but I left a note). I also moved ipython, which currently won't work for me, so I need to look more into this. I changed test.sh to test our contributions (ignoring ipython stuff)

As for guidance and what we test or not test, I think we should decide this later on experience (if we receive contributions).

tjhei:
Did you add the contributed_scripts/published/CHRU2014/*.pdf files on purpose?

sannecottaar:
Oops, no, let me fix that.

tjhei:
Thanks! Can you please try to squash your commits?

sannecottaar:
I could squash a couple. But I updated changes to ipython and then moved the files, and it is giving me issues with rebasing that. I ran into a sequence of error messages. Sorry, I'll squash more as I go.

tjhei:
I was able to squash all commits without any issues here:
```
git remote update
git checkout contributions_folder
git rebase -i upstream/master   # if "upstream" is the name of the official repo
```
then change all "pick" to "s" except the first. Done.

sannecottaar:
Thanks, I didn't know to use upstream/master, only tried with HEAD~x (which
also showed other people's commits).

tjhei:
So I can not make your ipython files work even with ipython3 installed on my machine. :-( Did you check if they work on cloud.sagemath.com?

sannecottaar:
I tried the old versions, and they work (but then run into the issues with
the new code). I guess something went wrong at some point when trying to
get them compatible with python3 and the testers. I'll make the changes
again to the old files and commit them in this pull request. Sorry for the
delay.

tjhei:
Can you please rename the folder to ``contrib``? That is a standard layout used in many projects and is a little more flexible.

tjhei:
I also don't see the "published" vs "education" as a sensible separation and would dump everything into ``contrib/`` directly.

tjhei:
Finally a question for you: should these files be part of the release .zip file?

sannecottaar:
Better? 
The ipynb are fixed too. 
I think these files could for now be part of the .zip.

bobmyhill:
Looks good to me, if Timo is happy! I think they should be part of the zipped release.

misc/readme.txt
```diff
@@ -0,0 +1,2 @@
+In case you are looking for any scripts referenced in Cottaar, Heister, Rose and Unterborn (2014), they are now located in:
+contributed_scripts/published/CHRU2014/
```
tjhei: this path is not updated

tjhei:
looks good otherwise.

tjhei:
Urgh, I can still not run the stupid ipython notebooks. I think my local installation is messed up. Sigh, python.

tjhei:
Sanne, I got them to work. I have a few changes I want to do, but lets move forward with this. Please change the path (see above) and then we can merge.

sannecottaar:
Done.

Issue #230, opened 2016-02-15
"Correct" dataset values


bobmyhill:
Fun fact: The values in the Stixrude and Lithgow-Bertelloni paper are not those used in hefesto or perplex. They are rounded severely enough that there is a noticeable discrepancy in transition pressure for some reactions (like wad-rw at 1673 K, for example, which is about 0.5 GPa off). I guess we missed this in benchmarking, as the minerals we chose to benchmark against were less affected by the rounding.

We should get the original dataset from somewhere. The PerpleX database has all the full values, so it would be a relatively easy task to write a mini program to convert stx11ver.dat to burnman-readable classes.

PR #231, opened 2016-02-16
Improve dataset accuracy

bobmyhill:
The values in Stixrude and Lithgow-Bertelloni are reported only to the number of significant figures for which they are confident. Annoyingly, this is not the number of significant figures required to reproduce the figures in their paper. This pull request fixes SLB_2011.py and adds appropriate tests, resolving Issue #230.


sannecottaar:
Wow, thanks for this!

lordkman:
Good catch Bob!

bobmyhill:
I also added an entropy term (via the new modify_properties interface) to a few SLB phases, which Stixrude and Lithgow-Bertelloni included (perhaps a bit obscurely in the paper) to account for magnetic/configurational entropy. This entropy term is unphysical at low temperatures (when phases are completely ordered and S should be zero).

bobmyhill:
Also also, if you were wondering why I've been banging on about approaching gibbs minimization from an equilibrium perspective, see the new benchmark in /misc/benchmarks/olivine_phase_diagram_benchmark.py ... it's much quicker than gridded minimization.

bobmyhill:
Ok, I see happy green lights!

tjhei:
What is the license for stx11ver.dat? I am not sure we can include it with the code...

bobmyhill:
stx11ver.dat is sourced from the Perple_X datasets directory; Perple_X doesn't have a licence.

I think it's fine to include here, but if you would rather it were removed, then I can just leave a readme...

tjhei:
> I think it's fine to include here

Do you have the right to license it under GPL2+? I don't think so. I would feel better if we don't include it. If I understand it correctly, you have added a script that loads it in an writes a mineral file we use? So, everything works without that file?

bobmyhill:
Good point; ok, I'm happy to remove it and leave a little note in the conversion file saying that the file is required, and from where it can be downloaded.

sannecottaar:
Why do we include the inversion file? I didn't include the ones that put
the seismic data in a consistent format.

bobmyhill:
There's also tc-ds62.txt in there, which also doesn't belong to us... I guess we should remove that too.

I like having the conversion files there for when new versions of the datasets are released (or for users to input their old copies of the datasets). If you think they should be removed, we can definitely do that, but then I take on the job of "Keeper of the dataset conversion scripts", and I don't like that idea.

tjhei:
> I like having the conversion files there

Agreed.

bobmyhill:
Ok, I've deleted the offending datasets, but added notes to the python scripts telling the user where to go to get the necessary files.

tjhei:
Thanks. We would need to squash the commits, otherwise the file will be in the history.

What are the  burnman/data/input_perplex/*dat files for?

bobmyhill:
I'm happy to squash the commits if you want, but it won't help... the HP dataset has been in burnman for over a year. Would you still prefer this PR squished?

The input_perplex files contain inputs to mineral benchmarks; they are snippets of code output from perplex. They're just tables of properties at given P and T.

tjhei:
Alright, looks good to me. Does anybody else have any comments?

sannecottaar:
Nope. Looks great!

PR #235, opened 2016-03-02
Changed excess_enthalpy references to excess_energy for clarity

bobmyhill:
A minor cosmetic tweak for clarity...

I was reminded recently of a point of contention wrt excess properties. It is conventional to describe excess properties in terms of an excess enthalpy, so I originally formulated the models in terms of excess enthalpy, entropy and volume at 0 K, 0 Pa. This is correct, but for the case of a model where Gex = a - bT + cP, it is clearer to use energy in place of enthalpy - then there's no need to declare the reference state, as G = E - TS + PV. It also makes the code a little easier to understand. See, for example, the excess_enthalpy function in solutionmodel.py.

As this PR changes the front-end naming convention for excess properties, the change should be included in Release 1.0.


bobmyhill:
Are we happy with this?

Issue #236, opened 2016-03-02
warning in test_solidsolution


tjhei:
I am getting
```
$ python test_solidsolution.py -v
test_1_gibbs (__main__.test_solidsolution) ... /home/heister/Dropbox/work/burnman-git/burnman/averaging_schemes.py:576: UserWarning: Oops, called reuss_average with Xi<=0!
  warnings.warn("Oops, called reuss_average with Xi<=0!")
```
@bobmyhill did you make the test? Is it correct to ignore this warning here?

PR #238, opened 2016-03-02
Shear modulus set to zero if minimum endmember value is 0.

bobmyhill:
Addresses Issue #236

### DIFF
--- a/burnman/solidsolution.py
+++ b/burnman/solidsolution.py
@@ -310,8 +310,10 @@ class SolidSolution(Mineral):
         Aliased with self.G
         """
         G_list = np.fromiter( (e[0].G for e in self.endmembers), dtype=np.float, count=self.n_endmembers)
-        return reuss_average_function( self.molar_fractions, G_list) 
-
+        G = 0.
+        if min(G_list) > 1.e-10:
+            G = reuss_average_function( self.molar_fractions, G_list) 
+        return G
     
     @material_property
     def p_wave_velocity(self):


tjhei:
What if the volume fraction is 0 for the minimum? Shouldn't you still average the other quantities?

bobmyhill:
I guess this is a question of what we want the software to do in unphysical situations. Only fluids have G = 0., and the user should never mix endmembers with undefined and defined Gs. 

Whatever we do, it should be consistent; if we throw an exception for the solid solution if G is not defined, we should do the same for the endmembers. At the moment, I've taken the conservative approach, and just set things to zero (like we currently have for the endmembers).

The alternative approach is to let the Reuss function only throw an exception for a negative argument.

tjhei:
My assumption is that a solid solution of 0% A and 100% B should behave as a B separately, no matter what A is.

bobmyhill:
Yes, that's true. Perhaps it would help if you tell me what you think should happen in the following situations:

1. The user asks for G for an endmember when the EoS doesn't define what it should be (e.g. for the hp_tmt EoS).
2. The user asks for G for a solid solution when none of the endmember EoSs define what G should be.
3. The user asks for G for a solid solution when all the molar fractions are non-zero and one endmember has an EoS with an undefined G.
4. The user asks for G for a solid solution when one endmember has an EoS with an undefined G and that endmember has a volume fraction of zero.

I would suggest that there should be one consistent answer to all of these questions. My preferred answer would be "throw exceptions". The code as it stands (including this PR) favours the alternative "return 0.", which I assume was originally chosen to allow the user to average Vp, Vs, rho simultaneously without worrying about NaNs or exceptions.

tjhei:
Good questions! We also have several places where we produce warnings...

bobmyhill:
So would you just like me to make the warning more verbose?

tjhei:
We could leave the logic as it currently is (warn and return 0), catch the warning in that one test, and call it a day?

bobmyhill:
Yeah, ok, that sounds fine ... do you want to do that?

Issue #240, opened 2016-03-06
Removing test output


bobmyhill:
After running tests, the directory structure is full of annoying output files. They can be removed with the line:
`rm -f */*.error */*/*.error */*-e */*/*-e contrib/tutorial/uncertainty.dat`

Can we write this into the test.sh script somehow? Maybe appending a flag if you want to keep them?
`test.sh --keep_output` would be nice.

tjhei:
What are these "*-e" files? I am not seeing them here.

bobmyhill:
The python files all have empty files with a .tmp-e suffix (e.g. misc/benchmarks/solidsolution_benchmarks.py.tmp-e) on both my mac and linux systems. 
Don't know why. 

In your PR, you don't remove contrib/tutorial/uncertainty.dat, but it is created by running test.sh. Can the tutorial be changed to run without producing output, perhaps?

bobmyhill:
Ooooh, it's a sed thing. sed -i'' -e is interpreted on my systems as sed -i'-e'.

tjhei:
huh, interesting!

tjhei:
does it work if you do "sed -i -e"? Does it work for you if you remove the "-e" (I have no idea why we do that)?

bobmyhill:
No, but in the sed manual it does warn the user not to leave the backup extension empty (corruption risk). Can I suggest adding a non-null extension, and then removing that files once during each loop?

e.g.
`sed -i'2' -e`
`rm $t.tmp2`

tjhei:
does ``sed -i '' -e ...`` work? (note the space)

> corruption risk

I don't think we care about corruption of some tmp file

bobmyhill:
Yes, adding the space works.

True, it's not doing any damage to the code if it becomes corrupted. I still don't think it's great practise, but if you're happy then I am too.

PR #241, opened 2016-03-06
remove .error files generated from test.sh

tjhei:
addresses #240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/contrib/tutorial/uncertainty.dat
 /misc/*.pdf
 /examples/example_woutput.txt
 /sphinx/_build/*
--- a/test.sh
+++ b/test.sh
@@ -43,13 +43,13 @@ then
 else
 
   sedthing="s#$fulldir#BURNMAN#g"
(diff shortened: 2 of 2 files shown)


tjhei:
updated.

bobmyhill:
Ok, great! Do you want to do something with uncertainty.dat?

tjhei:
I don't really like to make the tutorial more complicated. I now added the file to .gitignore. Acceptable?

tjhei:
wait with merging.

bobmyhill:
Is the output even necessary? Could the output just be suppressed?

tjhei:
step_3 writes the files and reads it in again when you want to plot. Of course we could add some magic that skips writing the file when we run the tests, but as I said, I think this is somewhat ugly for a tutorial.

tjhei:
I had to mess with the sed line even further because it stopped working on linux. Can you give the current version a try on mac os?

bobmyhill:
Yup, that's what I was suggesting before... It's the sed recommended style, so it should work in all systems. Ready to go?

tjhei:
working correctly under linux, so yes if it works for you.

PR #246, opened 2016-04-21
Gibbs minimizer

bobmyhill:
A working gibbs minimizer. This module includes a function which takes a bulk composition, an assemblage, and two constraints (fixed P, fixed T, or a compositional constraint), and calculates the phase fractions and compositions which minimize the Gibbs free energy.

To do list:
- [x] Add isochemical example
- [x] Add compositional example
- [x] Add lower mantle composition example
- [x] Change set_eqns so that the sum of phases is renormalised to one.
- [x] Implement solvus solving (where there can be more than one phase of the same type)
- [x] Change dictionary output so that variables and values are stored separately
- [x] Add solvus example
- [x] Create high level invariant and univariant functions
- [x] Add unit tests
- [x] Modify initial composition guesser to take zero phases into account.
- [x] Modify initial composition guesser for dependent endmember constraints
- [ ] Modify solver for dependent endmember constraints
- [x] Improve starting guesses / try multiple starting guesses?
- [ ] Fix constraints when endmembers are removed.


bobmyhill:
I've also added a higher level function which allows the bulk composition to vary. This is useful for binary phase diagrams, for example, where are interested in working out the bulk composition where a certain reaction takes place. The added degree of freedom means that one more P-T-X constraint is required.

tjhei:
Neat! Care to comment how you solved the problem you had before?

I wonder if it is better to have the ``gibbs_*`` functions return a tuple ``p,T,names`` instead of stuffing everything into a dictionary?

bobmyhill:
The two methods we tried last year had different problems. The reaction space solver that Ian set up was unable to deal with situations where amounts of individual phases were small (essentially a problem with round-off errors). The equilibrium solver that I set up involved an ad-hoc way of computing the bulk compositional constraints, which were sometimes incorrect.

The new solution uses the column space of the stoichiometric matrix to ensure that the bulk composition constraints are always fully satisfied. The left null space is used to ensure that the equilibrium relations are satisfied (giving us m equations for m endmembers). 

All other degrees of freedom (P, T, bulk composition) then need user-defined constraints (fixed P, T or some aspect of composition).

bobmyhill:
As for returning a tuple, I don't know - there are still a few more things to implement, and this may not be the best solution (especially when we start looking at solvi). We don't *need* to return anything, as the composite is updated with the correct conditions/compositions.

bobmyhill:
This PR has been superceded.

PR #250, opened 2016-11-25
Make endmember

bobmyhill:
Sometimes we want to make an endmember from a linear set of other endmembers, either as a guess when properties are not very well known, or for an ordered endmember which can be modelled as a simple mechanical mixture of endmembers (with an optional energy, entropy and volume increment). 

This PR creates a class (burnman.minerals.other.make_endmember()) that allows the user to define their own minerals.


burnman/minerals/other.py
```diff
@@ -12,10 +12,34 @@
...
+    For example, a crude approximation to a bridgmanite model might be
+    bdg = make_endmember([per, stv], [1.0, 1.0], [-15.e3, 0., 0.])
+
+    The endmember is actually a mechanical solid solution of constant composition, 
+    as this is the easiest way to create such a model. 
+    The function set_composition is called on initialisation, 
+    and shouldn't be called again.
+    """
+    def __init__(self, mineral_list, stoichiometry_list, ESV):
+        self.name = 'User-created endmember'
```
tjhei: Can you please document the constructor and explain the parameters? ``stoichiometry_list`` are the molar fractions in SolidSolution? Maybe move some of the documentation above to here?

burnman/minerals/other.py
```diff
@@ -12,10 +12,34 @@
 
 from .. import mineral_helpers as helpers
 from ..mineral import Mineral
+from ..solidsolution import SolidSolution
+from ..solutionmodel import *
 
 from .SLB_2011 import periclase, wuestite, mg_perovskite, fe_perovskite
 
-
+class make_endmember(SolidSolution):
```
tjhei: This not a typical naming for a class. Please use CamelCase and no underscore. The word "make" also sounds more like a function. Thinking about it, maybe you do want to create a function instead?
```
def make_endmember(mineral_list, stoichiometry_list, ESV):
  ss = SolidSolution(....)
  return ss
```

tjhei: also shouldn't this name convey that it is a ``mechanical solution``?

burnman/solutionmodel.py
```diff
@@ -159,7 +159,41 @@ def excess_entropy(self, pressure, temperature, molar_fractions):
...
+    """
+    An extremely simple class representing a mechanical solution model.
+    The ESV variable is a list containing a constant energy, 
+    entropy and volume offset from the linear mechanical solution.
+    """
+
+    def __init__(self, endmembers, ESV):
+        self.n_endmembers = len(endmembers)
+        self.formulas = [e[1] for e in endmembers]
+        self.ESV = ESV
```
tjhei: please assert that ESV has the correct size

burnman/solutionmodel.py
```diff
@@ -159,7 +159,41 @@ def excess_entropy(self, pressure, temperature, molar_fractions):
         """
         return 0.0
 
+class MechanicalSolution (SolutionModel):
 
+    """
+    An extremely simple class representing a mechanical solution model.
```
tjhei: can you give a definition what a mechanical solution is? Maybe in contrast to "ideal"?

burnman/minerals/other.py
bobmyhill: This is typical naming for a mineral or solid solution. See mg_fe_perovskite, for example. Do we want to change all the mineral names?!

As for naming to make it sound like a solution, I'm just using a mechanical solution as an easy way to address the problem without duplicating a whole bunch of code. This class is designed to create a fixed composition Material.

burnman/minerals/other.py
bobmyhill: Sure.

burnman/solutionmodel.py
bobmyhill: Will do.

burnman/solutionmodel.py
bobmyhill: Ok.

burnman/minerals/other.py
bobmyhill: The alternative is to write a new EoS that takes a set of minerals and adds together their properties. Perhaps it would be cleanest to write a dummy EoS that accesses all the functions in mechanical solution, allowing the class to be written as class MakeEndmember(Mineral):

A final possibility is to write a function make_endmember that returns an instance of class MadeEndmember(Mineral). I honestly don't know which alternative is best.

tjhei: > A final possibility is to write a function make_endmember that returns an instance of class MadeEndmember(Mineral).

I don't understand. Your class doesn't do anything SolidSolution doesn't already do. Why not return a ``SolidSolution``?

tjhei: > I don't want the user to be able to call set_composition.

But your class is derived from SolidSolution, so these function can be called! You are creating a class derived from SolidSolution so it inherits all functionality. Of course you can add/overwrite certain functionality but you don't do that right now.

burnman/solutionmodel.py
bobmyhill: Better?

burnman/mademember.py
```diff
@@ -0,0 +1,290 @@
...
+    All the solid solution parameters are expected to be in SI units.  This
+    means that the interaction parameters should be in J/mol, with the T
+    and P derivatives in J/K/mol and m^3/mol.
+    """
+
+    def __init__(self):
+        """
+        Set up matrices to speed up calculations for when P, T, X is defined.
+        """
+        if hasattr(self, 'mixture') == False:
```
tjhei: can you make ``mixture`` a parameter to this constructor instead?

burnman/mademember.py
```diff
@@ -0,0 +1,290 @@
...
+    """
+    assert(len(free_energy_adjustment) == 3)
+
+    class mechanical_mixture(SolidSolution):
+        def __init__(self, mineral_list, molar_amounts):
+            self.type = 'mechanical'
+            self.endmembers = [[m, ''] for m in mineral_list]
+            SolidSolution.__init__(self, molar_amounts)
+
+    class made_endmember(MadeMember):
```
tjhei: can't you do
```
x = MadeMember(mechanical_mixture(mineral_list, molar_amounts))
x.property_modifiers = ...
return x
```
this way you don't have to create a new class in the hierarchy that doesn't do anything useful. (This needs the change I mention above).

tjhei: or even better
```
x = MadeMember(mineral_list, molar_amounts)
x.property_modifiers = ...
return x
```
by adjusting the constructor to allow for that.

burnman/mademember.py
```diff
@@ -0,0 +1,290 @@
...
+from .solutionmodel import *
+from .averaging_schemes import reuss_average_function
+from .processchemistry import sum_formulae
+from . import constants
+
+
+class MadeMember(Mineral):
+
+    """
+    This is the base class for all solid solutions.
```
tjhei: do you now derive ``SolidSolution`` from ``MadeMember``? If yes, what is the difference between the two?

bobmyhill:
Oh, now I see what you were suggesting! Thanks!

In making the changes, I found one really annoying thing from my changes last year. The SolidSolution constructor doesn't work as a constructor outside a class, because it requires self.type and self.endmembers to already be defined. The obvious way round this would be to change the constructor to require type and endmembers as parameters for the constructor (along with all the other optional parameters), but this would break everyone else's code. 

Sorry, it was my suggestion to take these things out, and I'm only just starting to realise why they were there in the first place :/

Any suggestions?

bobmyhill:
For now, I've added optional parameters to __init__ (defaulting to None), and then added the following:
`        if solution_type is not None:
            self.type = solution_type
        else:
            if hasattr(self, 'type') == False:
                raise Exception(
                    "'type' attribute missing from solid solution")
        if endmembers is not None:
            self.endmembers = endmembers
        else:
            if hasattr(self, 'endmembers') == False:
                raise Exception(
                    "'endmembers' attribute missing from solid solution")`

burnman/mademember.py
bobmyhill: Great idea. Done!

burnman/mademember.py
bobmyhill: Yup.

burnman/mademember.py
bobmyhill: Ooops, forgot to update the blurb. It's still not finished, but I've removed the duplicate text now. No, SolidSolution is not derived from MadeMember.

tjhei:
> The SolidSolution constructor doesn't work as a constructor outside a class, because it requires self.type and self.endmembers to already be defined.

That makes it a pretty useless constructor. :-)

> The obvious way round this would be to change the constructor to require type and endmembers as parameters for the constructor

But that makes sense. You can always have different constructors or use optional parameters.

bobmyhill:
I think this one is good to go.

burnman/mademember.py
```diff
@@ -0,0 +1,274 @@
...
+        Aliased with self.C_v
+        """
+        return self.heat_capacity_p - self.molar_volume * self.temperature \
+            * self.thermal_expansivity * self.thermal_expansivity \
+            * self.isothermal_bulk_modulus
+
+
+def make_endmember(mineral_list, molar_amounts, free_energy_adjustment):
+    """
+    This function makes a new endmember from a linear combination of minerals
```
tjhei: please document arguments

burnman/mademember.py
```diff
@@ -0,0 +1,274 @@
...
+    This class is available as :class:`burnman.MadeMember`.
+
+    All the solid solution parameters are expected to be in SI units.  This
+    means that the interaction parameters should be in J/mol, with the T
+    and P derivatives in J/K/mol and m^3/mol.
+    """
+
+    def __init__(self, mineral_list, molar_amounts):
+        """
+        Set up matrices to speed up calculations for when P, T, X is defined.
```
tjhei: This comment looks out of place and doesn't document the constructor.

burnman/solidsolution.py
```diff
@@ -10,7 +10,9 @@
 
 from .mineral import Mineral, material_property
 from .solutionmodel import *
+from .processchemistry import sum_formulae
 from .averaging_schemes import reuss_average_function
+from .processchemistry import sum_formulae
```
tjhei: delete this line

burnman/mademember.py
tjhei: do you want to import this function in ``burnman/__init__``?

burnman/mademember.py
bobmyhill: Excellent point. Also a couple of the previous lines are no longer relevant after our reworking.

burnman/solidsolution.py
bobmyhill: Hah, good spot!

burnman/mademember.py
bobmyhill: Ok, and yes :)

bobmyhill:
Ok, all good.

tjhei:
Thanks. Somebody else have a comment about this PR?

burnman/mademember.py
```diff
@@ -0,0 +1,288 @@
...
+from .mineral import Mineral, material_property
+from .solidsolution import SolidSolution
+from .solutionmodel import *
+from .averaging_schemes import reuss_average_function
+from .processchemistry import sum_formulae
+from . import constants
+
+
+class MadeMember(Mineral):
+
```
sannecottaar: Sorry, I'm a bit late to the discussion on this. Why is it not named EndMember?

burnman/mademember.py
```diff
@@ -0,0 +1,288 @@
...
+    Returns
+    -------
+    m : Instance of MadeMember class
+        The desired mineral 
+    """
+    assert(len(free_energy_adjustment) == 3)
+
+    m = MadeMember(mineral_list, molar_amounts)
+
+    if free_energy_adjustment != []:
```
sannecottaar: Seems like this check is already asserted above. 
Why is this separate function needed? Most things I 'make' by initializing them, why is this constructed differently? If the only difference is that I want to be able to set them without free energy adjustments, this could just be an optional argument.

burnman/mademember.py
bobmyhill: This has caused confusion for both @sannecottaar and @tjhei, so I guess I should change the name. This class is for minerals made from a linear combination of existing minerals. Perhaps to be more obvious it should be called LinearCombinationMineral?

burnman/mademember.py
bobmyhill: Ooops. Yes,  I forgot to remove the assert, and now that you mention it, the function is indeed largely redundant. I did once have the energy increment in the constructor; I have no idea now why I took it out.

burnman/mademember.py
tjhei: CombinedMineral? MineralCombination?

bobmyhill: Oooh, I like CombinedMineral.

bobmyhill:
Whohoo, all good! Thanks @tjhei and @sannecottaar for all your help and advice, this looks so much better than my first attempt.

tjhei:
:+1:

PR #251, opened 2016-12-03
fixed solid solution constructor

bobmyhill:
constructor for SolidSolution now takes all arguments required to make the implemented SolutionModels. A class is no longer required to create a new SolidSolution instance.

Almost backwards compatible. The only change is that any calls to __init__ with only the molar_fractions parameter now need to name the variable, i.e.:

`burnman.SolidSolution.__init__(self, molar_fractions)`

should now read

`burnman.SolidSolution.__init__(self, molar_fractions=molar_fractions)`

### DIFF
--- a/burnman/solidsolution.py
+++ b/burnman/solidsolution.py
@@ -33,7 +33,14 @@ class SolidSolution(Mineral):
     and P derivatives in J/K/mol and m^3/mol.
     """
 
-    def __init__(self, molar_fractions=None):
+    def __init__(self,
+                 solution_type=None,
+                 endmembers=None,
+                 energy_interaction=None,
+                 volume_interaction=None,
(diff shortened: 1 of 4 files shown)


burnman/solidsolution.py
```diff
@@ -33,7 +33,14 @@ class SolidSolution(Mineral):
     and P derivatives in J/K/mol and m^3/mol.
     """
 
-    def __init__(self, molar_fractions=None):
+    def __init__(self,
```
tjhei: one option to make it compatible if you make molar_fractions the first argument. Thoughts?

tjhei:
looks good otherwise.

tjhei:
(as discussed it is better to keep it this way)

PR #267, opened 2016-12-21
Liquid iron equation of state

bobmyhill:
This PR provides the Anderson and Ahrens 1994 equation of state for liquid iron.


bobmyhill:
Ok, I think this is as good as it's gonna get. I had to make some choices to make the EoS self-consistent and as consistent as possible with the figures in the paper (not using the Grueneisen parameter off the principal isentrope, using a different expression for lambda). Maybe some bright spark will come along later and make further improvements, but for now I've documented the changes in the docstrings and labeled all the equations with the equation numbers from the paper.

misc/ref/liquid_iron_comparison.py.out
```diff
@@ -0,0 +1,27 @@
...
+Pressure (GPa), Temperature (K), Density (kg/m^3), Grueneisen parameter
+0.0 7019 1811.0 1.7350
+0.2 5500 3253.0 1.9377
+0.2 6000 2932.0 1.8832
+0.2 6500 2490.8 1.8214
+277.4 12643 5816.8 1.5675
+331.5 13015 7654.8 1.5862
+397.1 13417 9826.3 1.5877
+BURNMAN/burnman/tools.py: UserWarning: Warning, shear modulus is zero. If this mineral is not a liquid, then shear modulus calculations for the AA equation of state have not been implemented.
+  return func(*args)
```
tjhei: what is this "func(*args)" thing?

misc/ref/liquid_iron_comparison.py.out
```diff
@@ -0,0 +1,27 @@
...
+All EoS consistency constraints satisfied for 'burnman.minerals.other.liquid_iron'
+Pressure (GPa), Temperature (K), Density (kg/m^3), Grueneisen parameter
+0.0 7019 1811.0 1.7350
+0.2 5500 3253.0 1.9377
+0.2 6000 2932.0 1.8832
+0.2 6500 2490.8 1.8214
+277.4 12643 5816.8 1.5675
+331.5 13015 7654.8 1.5862
+397.1 13417 9826.3 1.5877
+BURNMAN/burnman/tools.py: UserWarning: Warning, shear modulus is zero. If this mineral is not a liquid, then shear modulus calculations for the AA equation of state have not been implemented.
```
tjhei: is this warning going to appear every time you look at a liquid? I don't like that.

misc/ref/table.py.out
```diff
@@ -424,14 +424,15 @@ SLB_2011_ZSB_2013.mg_perovskite                                     2.445e-05 2.
 SLB_2011_ZSB_2013.periclase                                         1.124e-05  1.61e+11      3.8      1.31e+11      2.1     0.0403 2   767.0         1.36  1.7     2.8
 SLB_2011_ZSB_2013.stishovite                                        1.402e-05  3.14e+11      3.8       2.2e+11      1.9     0.0601 3  1108.0         1.37  2.8     4.6
 SLB_2011_ZSB_2013.wuestite                                          1.226e-05  1.79e+11      4.9 59000000000.0      1.4     0.0718 2   454.0         1.53  1.7    -0.1
-Name                                                  V_0          K_0 Kprime_0           G_0 Gprime_0 molar_mass   n  Debye_0 grueneisen_0     q_0  eta_s_0
-other.Fe_Dewaele                                 6.75e-06    1.634e+11     5.38           nan      nan   0.055845   1
```
tjhei: why are the new AA minerals not appearing here?

misc/benchmarks/liquid_iron_comparison.py
```diff
@@ -0,0 +1,119 @@
...
+burnman.tools.check_eos_consistency(liq, P=10.e9, T=7000., tol=0.01, verbose=True)
+
+
+# Find heat capacities
+temperatures = np.linspace(1000., 15000., 101)
+Cvs = np.empty_like(temperatures)
+m = 0.055845
+rhos = np.empty_like(temperatures)
+densities = [5.e3,10.e3, 15.e3]
+#densities = [8.5e3, 9.56e3, 10.81e3, 12.28e3, 14.03e3, 16.14e3, 18.68e3]
```
tjhei: remove

burnman/eos/aa.py
```diff
@@ -0,0 +1,403 @@
...
+import numpy as np
+from scipy.optimize import brentq
+import warnings
+
+from . import equation_of_state as eos
+from ..constants import gas_constant
+
+class AA(eos.EquationOfState):
+    """
+    Base class for the liquid metal EOS detailed in :cite:`AA1994`.
```
tjhei: why "base class"? I don't see a place where you derive from it.

misc/ref/table.py.out
bobmyhill: They will do now. I've restructured table.py so that it can output minerals with different eoses from the same file.

misc/ref/liquid_iron_comparison.py.out
bobmyhill: No, only when you try to calculate the shear modulus. I think it's important to have a warning in the cases when the EoS has no implementation for the shear modulus. I've made the warning more explicit, with suggestions for the user.

misc/ref/liquid_iron_comparison.py.out
bobmyhill: That's a print out from the level 2 stack trace. I've changed the warning so that it doesn't appear any more.

bobmyhill:
@tjhei I can't reproduce the error message for table.py on my machine. Could you send me the file, and then I can track down the mystery. Thanks!

tjhei:
> @tjhei I can't reproduce the error message for table.py on my machine. Could you send me the file, and then I can track down the mystery. Thanks!

Huh. It looks like it only fails with python3.

tjhei:
Here is the file: 
[table.py.tmp.txt](https://github.com/geodynamics/burnman/files/1159916/table.py.tmp.txt)

It looks like the last three "other" blocks are in a different order when using python3.

tjhei:
You are depending on the order of items in a dictionary. Just sort them first.

bobmyhill:
Ooops. Thanks for catching that for me!

tjhei:
can you rebase to resolve the conflict in the .bib?

bobmyhill:
Apparently so :)

PR #269, opened 2017-01-17
Fix units of density in various examples

sannecottaar:
Apparently we think we can lift a cubic meter of mantle/core. I fixed this unit mistake in various places (unfortunately it is also in our paper).


tjhei:
ooops. ;-)

PR #271, opened 2017-01-17
Output the seismic synthetic codes

sannecottaar:
This adds the output files for axisem, and puts all functions for writing out within burnman. 
Let me know if you have questions/comments/suggestions. 

Here is an example of synthetics for a lower mantle replaced by some pyrolitic model compared to pure PREM (only down to 50 s).
prem_vs_pyrolite


examples/example_write_mineos_input.py
```diff
@@ -1,182 +0,0 @@
-# This file is part of BurnMan - a thermoelastic and thermodynamic toolkit for the Earth and Planetary Sciences
```
tjhei: why do you delete this file?

sannecottaar: I've merged the example into one example for both types of output example_writeout_for_synthetic_seismo.py (should have renamed this first and then adapted, but didn't know how to merge the files smartly afterwards)

burnman/output_seismo.py
```diff
@@ -0,0 +1,267 @@
...
+def write_axisem_input(rock, min_depth=670.e3, max_depth=2890.e3, T0= 1900,name='burnmantestrock',
+                       axisem_ref='axisem_prem_ani_noocean', plotting=False):
+    """
+    Writing velocities and densities to AXISEM (www.axisem.info) input file
+    Default is set to replacing the lower mantle with the BurnMan rock
+    Note:
+        - This implementation uses PREM to convert from depths to pressures to compute at
+        - This implementation assumes an adiabatic temperature profile, only T0 at min_depth can be set
+        - Currently, it only honors the discontinuities already in the synthetic input file, so it is best 
+            to only replace certain layers with burnman values (this should be improved in the future.
```
tjhei: missing closing paren.

burnman/output_seismo.py
```diff
@@ -0,0 +1,267 @@
...
+
+from . import tools
+from . import constants
+from . import seismic
+from . import geotherm
+
+def write_axisem_input(rock, min_depth=670.e3, max_depth=2890.e3, T0= 1900,name='burnmantestrock',
+                       axisem_ref='axisem_prem_ani_noocean', plotting=False):
+    """
+    Writing velocities and densities to AXISEM (www.axisem.info) input file
```
tjhei: Can you please document the arguments? Is ``burnmantestrock`` a good default? I would prefer if it was the full filename that is given here (including extension).

burnman/output_seismo.py
```diff
@@ -0,0 +1,267 @@
...
+    # WRITE OUT FILE
+    f = open('axisem_' + name + '.txt', 'w')
+    print('Writing axisem_' + name + '.txt ...')
+    f.write('# Input file for AXISEM created using BurnMan, replacing ' + axisem_ref+ ' with ' +name+ ' between ' +str(np.round(min_depth/1.e3)) + ' and ' + str(np.round(max_depth /1.e3)) +' km \n')
+    f.write('NAME ' + name + '\n')
+    for line in lines[2:18]:
+        f.write(line[:-1] + '\n')
+    for i in range(indrange[0]):
+        if i>0:
+            if ref_radius[i] ==ref_radius[i-1]:
```
tjhei: group the ifs together with ``and``

burnman/output_seismo.py
```diff
@@ -0,0 +1,267 @@
...
+                discontinuity = discontinuity + 1
+                f.write('#          Discontinuity   ' +str(discontinuity) + ', depth:    '+ str(np.round(ref_depth[i]/1.e3,decimals=2)) +' km \n')
+        f.write(
+            '%8.0f %9.2f %9.2f %9.2f %9.1f %9.1f %9.2f %9.2f %9.5f \n' %
+            (ref_radius[i],
+             ref_density[i],
+             ref_vpv[i],
+             ref_vsv[i],
+             ref_Qk[i],
```
tjhei: indentation? (here and a couple times below)

burnman/output_seismo.py
```diff
@@ -0,0 +1,267 @@
...
+        plt.plot(depthrange / 1.e3, rock_rho / 1.e3, color='r', linestyle='-',
+                 marker='o', markerfacecolor='r', markersize=1)
+
+        plt.title(axisem_ref + ' replaced with ' + name + ' between ' +
+                  str(min_depth / 1.e3) + ' and ' + str(max_depth / 1.e3) + ' km')
+        plt.legend(loc='lower right')
+        plt.show()
+
+
+def write_mineos_input(rock, min_depth=670.e3, max_depth=2890.e3, name='burnmantestrock',
```
tjhei: same ``filename`` discussion as above

examples/example_writeout_for_synthetic_seismo.py
```diff
@@ -0,0 +1,73 @@
+# This file is part of BurnMan - a thermoelastic and thermodynamic toolkit for the Earth and Planetary Sciences
+# Copyright (C) 2012 - 2015 by the BurnMan team, released under the GNU
+# GPL v2 or later.
+"""
+This example script produces an input file to compute normal modes in
```
tjhei: add axisem stuff here please

sannecottaar:
I don't get this error on my computer. Anybody have a clue?
!  example_writeout_for_synthetic_seismo.py ... FAIL
Traceback (most recent call last):
  File "<stdin>", line 7, in <module>
  File "example_writeout_for_synthetic_seismo.py", line 71, in <module>
    burnman.output_seismo.write_axisem_input(pyrolitic_mantle, filename='axisem_pyrolite.txt', plotting=True)
  File "/main/burnman/burnman/output_seismo.py", line 62, in write_axisem_input
    ref_radius = table[:, 0]
IndexError: too many indices

tjhei:
> I don't get this error on my computer. Anybody have a clue?

input file missing?

sannecottaar:
It's there, unless it isn't finding it. But I tried to use the same setup as read_table in tools.py uses (using pkgutil)

burnman/output_seismo.py
```diff
@@ -0,0 +1,279 @@
...
+    
+    # Load reference input
+    datastream = pkgutil.get_data('burnman', 'data/input_seismic/' + axisem_ref)
+    lines = [line.strip()
+                 for line in datastream.decode('ascii').split('\n') if line.strip()]
+    table = []
+    for line in lines[18:]:
+        numbers = np.fromstring(line, sep=' ')
+        if len(numbers)>0:
+            table.append(numbers)
```
tjhei: I am getting some rows that just contain the number "-1", so this thing can not be converted into an array below. I assume you want to skip those rows?

tjhei:
That seemed to have worked (if you adjust the output)

burnman/output_seismo.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+
+
+    
+    # Load reference input
+    datastream = pkgutil.get_data('burnman', 'data/input_seismic/' + axisem_ref)
+    lines = [line.strip()
+                 for line in datastream.decode('ascii').split('\n') if line.strip()]
+    table = []
+    for line in lines[18:]:
```
tjhei: I don't particularly like skipping exactly 17 lines, because it is not very robust. You could for example skip every line until you encounter one that starts with ``COLUMNS``.

sannecottaar:
Yes, I was thinking it would be a lot more elegant to form a generic table with a rock and geotherm (much more modular) and then input that into the routines to make output. The reference files are a bit fixed the way they are now, thinking about it now, I could implement that it uses any 1D seismic model stored in BurnMan as a background model (would have to include anisotropy in that though). This implementation would also be best after the models have been changed into 'generic tables'.

PR #276, opened 2017-02-18
Implemented perplex-output parsing tools and material model

bobmyhill:
PerpleX is an extremely powerful and commonly used tool for creating equilibrium phase diagrams. 

This PR request includes files which:
- call the werami software bundled with PerpleX to create a table file which contains thermodynamic data
- implement a PerplexMaterial class (derived from Material), which reads in a table file and can then be interrogated as for any other material model.


bobmyhill:
Requires scipy 0.14+

misc/create_burnman_readable_perplex_table.py
```diff
@@ -0,0 +1,86 @@
+from subprocess import Popen, PIPE, STDOUT
+import argparse
+
+'''  
+This python file uses PerpleX's werami software
```
tjhei: so this is a command line utility, right? Would it make sense to have the functionality below be a python function in burnman/perplex.py and call it from this file? Or is it not useful to call this programmatically?

bobmyhill: werami is quite slow, so this is certainly not something I would use frequently, but you're right. 
I'll add a function to perplex.py.

tjhei:
> Requires scipy 0.14+, otherwise RegularGridInterpolator is missing.

jup, that is not ideal. Ubuntu 14.04 ships with 0.13 for example (that is what I am using for the tester). What do you propose we do about this?

bobmyhill:
Is it ok to import the grid interpolator from scipy? I've added a comment that that is where it's from...

tjhei:
can you add create_burnman_readable_perplex_table.py to the list of exceptions in test.sh so it is skipped?

tjhei:
> Is it ok to import the grid interpolator from scipy?

No, that won't work because their source license doesn't allow that.

bobmyhill:
Ok, there isn't much of a hit to use interp2d instead. 
All fixed.

burnman/perplex.py
```diff
@@ -0,0 +1,294 @@
...
+    stdout = p.communicate(input=stdin)[0]
+    print(stdout)
+    out = [s for s in stdout.split('\n') if "Output has been written to the" in s][0].split()[-1]
+    rename(out, outfile)
+    print('Output file renamed to {0:s}'.format(outfile))
+    print('Processing complete')
+
+    
+def read_2D_perplex_file(filename):
+    with open(filename, 'r') as f:
```
tjhei: this is an internal function, right? Maybe make this a member of the PerPlexMaterial or at least start with "_"?

bobmyhill: Good point. I've done both. Should be ready to roll.

tjhei:
is this ready to get merged?

bobmyhill:
It is for me!

PR #277, opened 2017-02-19
Evaluate now takes an n dimensional list of pressures and temperatures

bobmyhill:
Useful when using np.meshgrid to create a grid of properties.

### DIFF
--- a/burnman/material.py
+++ b/burnman/material.py
@@ -176,10 +176,10 @@ class Material(object):
         ----------
         vars_list : list of strings
             Variables to be returned for given conditions
-        pressure : array of float
-            Array of pressures in [Pa].
-        temperature : float
-            Array of temperatures in [K].
+        pressures : ndlist or ndarray of float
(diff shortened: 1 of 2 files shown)


tjhei:
I am not sure I understand, you want to allow any shape for the input and then return the output in the same shape? Not sure if this is a great idea. We would need to at least check that the shapes of p and T are compatible.

sannecottaar:
Seems reasonable to do a grid at once. I think we always needed to put the same shape in (couldn't calculate a range of temperatures at one pressure). Check would be useful.

tjhei:
I am fine with this, provided:
1. we assert that the shapes you put in are compatible
2. you add unit tests that check that if you put in a list, a 1d array, and a 2d array and verify that the output shape is correct.

bobmyhill:
Not sure why this is failing ... perplex.py isn't even in this PR

bobmyhill:
Ok, all fixed.

tjhei:
great!

PR #278, opened 2017-02-19
Smooth gridded property

bobmyhill:
This PR adds a gaussian smoothing function, which operates on a rock and returns a smoothed property grid. This PR will be useful for input to geodynamic simulations, which have trouble handling strong gradients in entropy and volume.

Uses the commits in PR #276


bobmyhill:
Now also contains the smoothing functions required for geodynamic applications, and an example demonstrating that functionality.

burnman/tools.py
```diff
@@ -643,3 +644,109 @@ def check_eos_consistency(m, P=1.e9, T=300., tol=0.01, verbose=False):
...
+    temperature_resolution = temperatures[1] - temperatures[0]
+
+    sigma = (temperature_stdev/temperature_resolution,
+             pressure_stdev/pressure_resolution)
+
+    padding = (int(np.ceil(truncate*sigma[0])), int(np.ceil(truncate*sigma[1])))
+    padded_property_grid = _pad_ndarray_inverse_mirror(property_grid, padding)
+
+    smoothed_padded_property_grid = gaussian_filter(padded_property_grid,
+                                                    sigma=sigma)
```
tjhei: do you need to pass ``truncate`` here?

burnman/tools.py
```diff
@@ -643,3 +644,109 @@ def check_eos_consistency(m, P=1.e9, T=300., tol=0.01, verbose=False):
...
+        meshed temperatures
+    property_grid: 2D numpy array
+        meshed property grid
+    smoothed_property_grid: 2D numpy array
+        meshed smoothed property grid
+
+    """
+    
+    pp, TT = np.meshgrid(pressures, temperatures)
+    property_grid = material.evaluate([gridded_property], pp, TT)[0]
```
tjhei: I wonder if it makes sense to instead pass the ``property_grid`` into this function and do the evaluation of the material outside. This function could operate on a 2d grid or a 1d table...

burnman/tools.py
```diff
@@ -643,3 +644,109 @@ def check_eos_consistency(m, P=1.e9, T=300., tol=0.01, verbose=False):
             print('Not satisfied all EoS consistency constraints for {0:s}'.format(m.to_string()))
             
     return consistency
+
+
+def _pad_ndarray_inverse_mirror(array, padding):
+    """
+    Pads an ndarray according to an inverse mirror
```
tjhei: You saw the different ``mode`` options of ``gaussian_filter``, right? I can not find their precise definition, but did you try  ``mode=refect``?

burnman/tools.py
```diff
@@ -643,3 +644,109 @@ def check_eos_consistency(m, P=1.e9, T=300., tol=0.01, verbose=False):
             print('Not satisfied all EoS consistency constraints for {0:s}'.format(m.to_string()))
             
     return consistency
+
+
+def _pad_ndarray_inverse_mirror(array, padding):
```
tjhei: wouldn't this function be easier, if you execute this for each dimension separately?

burnman/tools.py
tjhei: The reason is that one might filter a profile that comes from some other program in a similar way.

burnman/tools.py
```diff
@@ -643,3 +645,124 @@ def check_eos_consistency(m, P=1.e9, T=300., tol=0.01, verbose=False):
...
+        padding = (int(np.ceil(truncate*sigma[0])), int(np.ceil(truncate*sigma[1])))
+        padded_property_grid = _pad_ndarray_inverse_mirror(property_grid, padding)
+        
+        smoothed_padded_property_grid = gaussian_filter(padded_property_grid,
+                                                        sigma=sigma)
+    else:
+        padded_property_grid = property_grid
+        
+        smoothed_padded_property_grid = gaussian_filter(padded_property_grid,
+                                                        sigma=sigma, mode=mode)
```
tjhei: do you need to pass ``truncate`` too?

burnman/tools.py
```diff
@@ -643,3 +645,124 @@ def check_eos_consistency(m, P=1.e9, T=300., tol=0.01, verbose=False):
...
+
+    for i, idx in enumerate(padded_indices):
+        padded_array[idx] = 2.*padded_array[edge_indices[i]] - padded_array[mirror_indices[i]]
+
+    return padded_array
+
+
+def interp_smoothed_property(material, gridded_property, pressures, temperatures,
+                             pressure_stdev=0, temperature_stdev=0, truncate=4.0,
+                             mode='inverse_mirror'):
```
tjhei: document ``mode``

tjhei:
Sorry, I looked at a subset of your changes where you didn't use the other modes through ``gaussian_filter`` yet. Ignore that comment.

tjhei:
I guess the main question I have is:
Can we change this function to work on 1d or 2d arrays already computed (for example through ``material.evaluate`` like in your example). This would make it more flexible (apply to more than one column, apply to data coming out of planet builder, apply to data loaded in, etc.).

burnman/tools.py
bobmyhill: I don't think so. There's padding in the corners that needs to be dealt with as well.

burnman/tools.py
bobmyhill: reflect and mirror are essentially the same; the boundaries are treated as follows:
           pad | array
reflect: 4 3 | 3 4 6
mirror:  6 4 | 3 4 6
newfn:  0 2 | 3 4 6

burnman/tools.py
bobmyhill: Yeah, I was wondering about this as I was working on it. You're right, probably better to generalise and have a less generic higher level function if necessary.

burnman/tools.py
bobmyhill: Might be a good plan!

burnman/tools.py
bobmyhill: Ooops, definitely!

burnman/tools.py
tjhei: I don't think you need padding in the corners or at least the values are never used. This is because the filter will run in each 1d direction separately.

bobmyhill: It's true that the filter runs along each axis separately, but that means that the corners *are* used. Take this example, where there is a 1 cell padding all the way round the array and I take the filter to be (0.25*(values either side) + 0.5*value):

1   2 2   1             7/4   7/4 
1 | 3 3 | 1  -->   | 10/4  10/4  |  ->  | 10/4  10/4 |
1 | 4 4 | 1         |  13/4  13/4 |       | 13/4   13/4 |
1   5 5   1            16/4  16/4 

The corner padding is used in the first step.

bobmyhill: Well that wasn't well formatted...
Starting array:

| 1 | 2 | 2 | 1 |
|---|---|---|---|
| 1 | 3 | 3 | 1 |
| 1 | 4 | 4 | 1 |
|---|---|---|---|
| 1 | 5 | 5 | 1 |

bobmyhill: Well, that wasn't well formatted. Starting array:

| 1 | 2 | 2 | 1 |
| 1 | 3 | 3 | 1 |
| 1 | 4 | 4 | 1 |
| 1 | 5 | 5 | 1 |

After first step (note the use of the corner values)

| - | 7/4  | 7/4  | - |
| - | 10/4 | 10/4 | - |
| - | 13/4 | 13/4 | - |
| - | 16/4 | 16/4 | - |

bobmyhill: Final array:

| - | -    | -    | - |
|---|------|------|---|
| - | 10/4 | 10/4 | - |
| - | 13/4 | 13/4 | - |
| - | -    | -    | - |

bobmyhill:
Ok, what do you think of this?
smoothed_array is basically an alias of gaussian_filter, but with the new padding scheme implemented. This can be the generic function for pre-computed arrays, and will work in n dimensions. 

The new function uses smoothed_array and interp2d for 2D arrays, and so is useful for the specific case of smoothing and sampling 2D P-T grids.

burnman/tools.py
tjhei: I thought extension & filter is applied in each coordinate direction separately, at least in gaussian_filter, so:
start:
```
3 3
4 4
```
extend in x direction:
```
1 3 3 1
1 4 4 1
```
apply in x direction:
```
5/2 5/2
13/4 13/4
```
extend in y direction (I made the numbers up):
```
1 1
5/2 5/2
13/4 13/4
4 4
```
finally apply in y direction.

bobmyhill: True, that's how gaussian_filter works, but in order to add a new "mode" (which is necessary for our data) I would have to write my own version of gaussian_filter and use gaussian1d. I felt that my current strategy (using scipy's function rather than rewriting it) was slightly more sensible, but I can certainly rewrite gaussian_filter if you want? Processing would certainly be faster.

bobmyhill: Looking a bit more carefully, If I really wanted to add a mode I have to burrow down through scipy:
gaussian_filter -> gaussian_filter1d -> correlate1d. I append my previous statement by saying I'd really rather not do that :/

tjhei: The only thing that you would have to do is calling ``gaussian_filter1d`` for each dimension and extend the array using your own method beforehand or not?

bobmyhill:
Well, yes, but why would we want to copy the big (unpadded) array n times
for an n dimensional array?  I'm not sure the change will be any faster,
and it won't be much more readable. Happy to do it if there's a good reason.

tjhei:
> Well, yes, but why would we want to copy the big (unpadded) array n times
> for an n dimensional array?  I'm not sure the change will be any faster,
> and it won't be much more readable. Happy to do it if there's a good reason.

Yeah, if you care about speed than your solution is probably better. I don't feel strongly about it.

burnman/tools.py
```diff
@@ -643,3 +645,165 @@ def check_eos_consistency(m, P=1.e9, T=300., tol=0.01, verbose=False):
...
+    """
+    Creates a smoothed array on a regular 2D grid. Smoothing 
+    is achieved using burnman.tools.smooth_array(). 
+    Outputs scipy.interpolate.interp2d() interpolators 
+    which can be used to query the array, or its derivatives in the 
+    x- and y- directions.
+
+    Parameters
+    ----------
+    material : burnman Material
```
tjhei: please remove

burnman/tools.py
```diff
@@ -643,3 +645,165 @@ def check_eos_consistency(m, P=1.e9, T=300., tol=0.01, verbose=False):
...
+        smoothed_array = smoothed_padded_array[slices]
+    else:
+        smoothed_array = gaussian_filter(array, sigma=sigma, mode=mode)
+        
+    return smoothed_array
+
+
+def interp_smoothed_array_and_derivatives(array,
+                                          x_values, y_values,
+                                          x_stdev=0, y_stdev=0,
```
tjhei: why 0 as a default?

tjhei:
This looks good to me. Can you please add a unit test that applies it to something like
``
1 2 
3 4
``
?

burnman/tools.py
bobmyhill: Sure!

burnman/tools.py
bobmyhill: Dunno. Seemed the only objective choice (apart from 1 grid spacing).

tjhei:
okay, this is fine with me now. Anybody else have comments?

tjhei:
(except the failing test)

bobmyhill:
Ok, that example is fixed.
It worked fine on my machine; scipy has hugely improved its ability to work on n-dimensional arrays recently (I have version 0.18.1), but it has rather screwed up backwards compatibility!

I'm happy with this one now :)

PR #280, opened 2017-03-02
Improved heat capacity accuracy, consistency checks

bobmyhill:
The Holland and Powell equation of state lacked a term in the derivation of heat capacity, which caused a slight deviation from analytical calculations at high pressures.

This term has a minor effect on calculated heat capacities (<1 J/K/mol at 100 GPa), but for consistency it is now included.

The correction allows us to be more rigorous in our EoS checking, so I've also added a few lines to the check_eos_consistency function.


bobmyhill:
Finally, a compact PR! Should make a nice change from those 2000 line monsters :) 
Think this one is ready now.

PR #284, opened 2017-03-28
added mineral data set from Kurnosov et al. 2017

sannecottaar:


burnman/minerals/KMFBZ_2017.py
```diff
@@ -0,0 +1,254 @@
...
+
+class ferropericlase(SolidSolution):
+
+    def __init__(self, molar_fractions=None):
+        self.name = 'magnesiowustite/ferropericlase'
+        self.type = 'symmetric'
+        self.endmembers = [[periclase(), '[Mg]O'], [wuestite(), '[Fe]O']]
+        self.energy_interaction = [[13.e3]]
+
+        SolidSolution.__init__(self, molar_fractions=molar_fractions)
```
bobmyhill: suggest just SolidSolution.__init__(self, molar_fractions), same as perovskite?

burnman/minerals/KMFBZ_2017.py
```diff
@@ -0,0 +1,254 @@
...
+
+class bridgmanite(SolidSolution):
+
+    def __init__(self, molar_fractions=None):
+        self.endmembers = [[mg_perovskite(), '[Mg][Si]O3'],
+            [fe_perovskite(), '[Fe][Si]O3'],
+            [al_al_perovskite(), '[Al][Al]SiO3'],
+            [fe_al_o3(), '[Fe][Al]SiO3'],]
+        self.type = 'ideal'
+        SolidSolution.__init__(self, molar_fractions)
```
bobmyhill: This class should ideally have a name...

burnman/minerals/KMFBZ_2017.py
```diff
@@ -0,0 +1,254 @@
...
+        formula = dictionarize_formula(formula)
+        self.params = {
+            'name': 'Mg_Perovskite',
+            'formula': formula,
+            'equation_of_state': 'slb3',
+            'F_0': -1368000.0,
+            'V_0': 2.4445e-05,
+            'K_0': 2.51e+11,
+            'Kprime_0': 4.1,
+            'Debye_0': 905,
```
bobmyhill: float?

burnman/minerals/KMFBZ_2017.py
```diff
@@ -0,0 +1,254 @@
...
+        formula = dictionarize_formula(formula)
+        self.params = {
+            'name': 'Fe_Perovskite',
+            'formula': formula,
+            'equation_of_state': 'slb3',
+            'F_0': -1043000.0,
+            'V_0': 2.534e-05,
+            'K_0': 2.72e+11,
+            'Kprime_0': 4.1,
+            'Debye_0': 871,
```
bobmyhill: float (and the rest)...

burnman/minerals/KMFBZ_2017.py
```diff
@@ -0,0 +1,254 @@
+# This file is part of BurnMan - a thermoelastic and thermodynamic toolkit for the Earth and Planetary Sciences
+# Copyright (C) 2012 - 2015 by the BurnMan team, released under the GNU
+# GPL v2 or later.
+
+
+"""
+from Kurnosov, Marquardt, Frost, Ballaran and Ziberna, Nature 2017
```
bobmyhill: add title, doi?

burnman/minerals/KMFBZ_2017.py
tjhei: what do you mean? This is just a call to the base class constructor.

sannecottaar:
It would be nice to have a more elegant solution for the init file. Maybe we could start all the filenames with mindatabase_, and then import all mindatabase_*.py or so...

tjhei:
> It would be nice to have a more elegant solution for the init file. Maybe we could start all the filenames with mindatabase_, and then import all mindatabase_*.py or so...

There is no easy way to do this (*) and it is not that much work, right?

*: See http://stackoverflow.com/questions/1057431/loading-all-modules-in-a-folder-in-python if you don't believe me.

sannecottaar:
glob can do it. It is not much work, but it is an issue that people would/could run into and loose time on if they do not know python that well.

bobmyhill:
I'm keen on the globbing idea, although maybe something for another PR? @sannecottaar, are you happy with this now? I don't see any problems with merging.

tjhei:
> I'm keen on the globbing idea, although maybe something for another PR?

I am not, but definitely a separate PR, yes.

sannecottaar:
Yes, this can be merged.

PR #286, opened 2017-03-29
Anisotropy

bobmyhill:
All minerals are anisotropic. An increasing number of experimental investigations are constraining individual elastic constants for Earth materials, and it won't be long before equations of state like SLB are used without using isotropic approximations.

This PR introduces the main functions required to turn a set of Cijs into observable elastic properties. An example is included.


burnman/anisotropy.py
```diff
@@ -0,0 +1,449 @@
...
+# Copyright (C) 2012 - 2015 by the BurnMan team, released under the GNU
+# GPL v2 or later.
+
+from __future__ import absolute_import
+import numpy as np
+import matplotlib.pyplot as plt
+
+class AnisotropicMaterial(object):
+    """
+    Base class for anisotropic materials
```
tjhei: Is this really a base class? Below you are using it directly instead of creating new classes.

tjhei: I guess (for now?) ``Material`` and ``AnisotropicMaterial`` are completely unrelated?

burnman/anisotropy.py
```diff
@@ -0,0 +1,449 @@
...
+            i, j = self._voigt_index_to_ij(m)
+            for n in range(6):
+                k, l = self._voigt_index_to_ij(n)
+                stiffness_tensor[i][j][k][l] = voigt_notation[m][n]
+                stiffness_tensor[j][i][k][l] = voigt_notation[m][n]
+                stiffness_tensor[i][j][l][k] = voigt_notation[m][n]
+                stiffness_tensor[j][i][l][k] = voigt_notation[m][n]
+        return stiffness_tensor
+    
+    @property
```
tjhei: can you explain to me what this decorator here does?

burnman/anisotropy.py
```diff
@@ -0,0 +1,449 @@
...
+            j=2
+        elif m == 5:
+            i=0
+            j=1
+        else:
+            i=m
+            j=m
+        return i, j
+
+    def _normalize(self, vector):
```
tjhei: maybe a global function instead of a member of this class?

burnman/anisotropy.py
```diff
@@ -0,0 +1,449 @@
...
+        return C
+        
+    def _voigt_index_to_ij(self, m):
+        """
+        Returns the ij (or kl) indices of the 
+        stiffness tensor which correspond to those 
+        of the Voigt notation m (or n).
+        """
+        if m == 3:
+            i=1
```
tjhei: just ``return 1,2`` directly (same below)

burnman/anisotropy.py
sannecottaar: Should they be @material_property?

burnman/anisotropy.py
bobmyhill: @tjhei Good point, not a base class yet. I've added a couple of sentences which should clarify my intentions for the class. 

At the moment, Material and AnisotropicMaterial are unrelated, but they will probably become more closely related with time.

burnman/anisotropy.py
bobmyhill: Thanks!

burnman/anisotropy.py
bobmyhill: Done. I needed to restructure tools.py to avoid circular imports, but I think it's better now anyway.

burnman/anisotropy.py
bobmyhill: It means that material.bulk_modulus_voigt is only calculated when needed, and doesn't need to be suffixed with ().

tjhei: The first part is incorrect. ``@property`` will always call this function and recompute the result. I wrote ``@material_property`` that caches the value but it invalidates the result when you call ``set_state``, which you don't have here.

bobmyhill: @tjhei I think there are some communication issues here. I meant that unless I ask for material.bulk_modulus_voigt, it won't be calculated. Is that correct? 

@tjhei and @sannecottaar , the points you make about caching, and more generally deriving from Material are good ones. I've reworked AnisotropicMaterial to derive from Material. Eventually, it would be good to add a set_state, but there isn't any point at the moment, as we don't have an appropriate equation of state to use. (Coding up the full SLB EoS might be a nice project to give to a student, but is way outside the scope of this PR).

tjhei: >  I meant that unless I ask for material.bulk_modulus_voigt, it won't be calculated. Is that correct?

Of course, but that has nothing to do with ``@property``.

tjhei:
> I've reworked AnisotropicMaterial to derive from Material. Eventually, it would be good to add a set_state, but there isn't any point at the moment, as we don't have an appropriate equation of state to use.

When using ``@material_property`` you need to be careful that you need to call ``self.reset()`` whenever any state of the object changes that may influence any of the cached properties. This is done in ``Material.set_state()`` for example.

bobmyhill:
@tjhei sure. At the moment, this is a static material, so we don't need a set_state. Now that I've reworked the class, it should be easy enough to modify (for example, if we want to make an equation of state with Cij' and Cij'')

bobmyhill:
Ok, I've made all the changes you suggested (plus added a unit test) ... anything else @tjhei?

burnman/anisotropy.py
```diff
@@ -0,0 +1,471 @@
+# This file is part of BurnMan - a thermoelastic and thermodynamic toolkit for the Earth and Planetary Sciences
+# Copyright (C) 2012 - 2015 by the BurnMan team, released under the GNU
```
ian-r-rose: We should update the copyright year.

burnman/anisotropy.py
```diff
@@ -0,0 +1,471 @@
...
+        eigenvalues, eigenvectors = np.linalg.eig(Tik)
+
+        idx = eigenvalues.argsort()[::-1]   
+        eigenvalues = np.real(eigenvalues[idx])        
+        eigenvectors = eigenvectors[:,idx]
+        velocities = np.sqrt(eigenvalues/self.rho)
+
+        return velocities, eigenvectors
+
+    def plot_velocities(self):
```
ian-r-rose: Should we have plotting functions within `Mineral` classes? This seems like a new pattern.

burnman/anisotropy.py
```diff
@@ -0,0 +1,471 @@
...
+            
+            ticks = np.linspace(tmin, tmax, nt)
+            im.append(ax[i].contourf(theta, r, item, ndivs, cmap=plt.cm.jet_r, vmin=vmin, vmax=vmax))
+            lines = ax[i].contour(theta, r, item, ticks, colors=('black',), linewidths=(1,))
+            
+            cbar = fig.colorbar(im[i], ax=ax[i], ticks=ticks)
+            cbar.add_lines(lines)
+
+        plt.tight_layout()
+        plt.show()
```
ian-r-rose: I would at least suggest returning the matplotlib `artist` objects, so users could change the plot or incorporate them into other plots.

burnman/tools.py
```diff
@@ -50,6 +49,12 @@ def flatten(l): return flatten(l[0]) + (flatten(l[1:]) if len(l) > 1 else []) if
 def round_to_n(x, xerr, n):
     return round(x, -int(np.floor(np.log10(np.abs(xerr)))) + (n - 1))
 
+def normalize(a, order=2, axis=-1):
```
ian-r-rose: Perhaps name this `unit_normalize`?

examples/example_anisotropy.py
```diff
@@ -0,0 +1,103 @@
...
+
+if __name__ == "__main__":
+    # Let's first look at an isotropic material
+    # As an example, let's use the lambda (C12), mu (C44) and rho
+    # for flow basalt given by Christensen et al. (1980)
+    # Initial Reports of the Deep Sea Drilling Project 59: 515-17.
+
+    elastic_constants = [0.4e11, 0.24e11]
+    rho = 2.735e3
+    basalt = anisotropy.AnisotropicMaterial(elastic_constants, rho, 'isotropic')
```
ian-r-rose: This is one place where subclassing might be clearer (making an isotropic `AnisotropicMaterial`). No strong objections, though.

burnman/anisotropy.py
bobmyhill: Done.

burnman/anisotropy.py
bobmyhill: Good point... probably not.

burnman/anisotropy.py
bobmyhill: I've moved the function to the example - it's probably more appropriate than a fixed function.

burnman/tools.py
bobmyhill: Done.

burnman/anisotropy.py
ian-r-rose: Timo and I just looked, and we have it in a couple of places at the moment.

However, the reason we just had to look was that matplotlib was error-ing due to a bad assumption about the backend. To me that seems like a good reason not to assume a particular plotting environment. At least not in the core that is imported at the top.

burnman/anisotropy.py
ian-r-rose: Okay, sounds good. We should probably move a couple of other plotting functions out of the core.
Perhaps a plotting submodule that is not automatically imported?

bobmyhill: Sounds like a plan... I think we're going to need the plotting routines more and more now that we're developing high-level tools. For another PR though, there's a lot in here already!

burnman/anisotropy.py
```diff
@@ -0,0 +1,455 @@
...
+from __future__ import absolute_import
+from __future__ import print_function
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+from .tools import unit_normalize
+from .material import Material, material_property
+
+class AnisotropicMaterial(Material):
```
ian-r-rose: Do we get any value from deriving from Material here?

bobmyhill: At the moment, we get very little benefit, but we will eventually get a lot out of this, when we have an EoS to describe how Cijs change with pressure and temperature. I think it's valuable to prepare for this now.

bobmyhill: I've just checked all of the variable names, and they are suitable.

PR #290, opened 2017-05-07
Improve smoothing for geodynamic adiabat, create ipython notebook

bobmyhill:
When smoothing entropy and volume grids (to compute smoothed adiabatic profiles), it is important that the temperature smoothing gaussian width is wide enough to capture the changes in slope of the isentrope. A simple way to do this is implemented here. 

To make incorporation into an ipython notebook trivial, I've also moved all of the inputs to the beginning of the file


bobmyhill:
@tjhei Here's a notebook for you!

PR #296, opened 2017-06-21
define geotherms in depth

sannecottaar:
Currently some of our geotherms relied on a pressure to depth conversion using PREM within the routine. This is not elegant in the case that people want to do things self-consistently and use the pressures from their rock. 
This change means that geotherms can both be defined in depth and pressure (for the adiabat), which I think is fine and correct. The change only required very minor changes to the examples.


sannecottaar:
This also is a setup to include linear geotherms, ways to add boundary layers, etc. in geotherm.py later.

bobmyhill:
You've got a few failing tests. Otherwise, this looks good. 

I've been wondering for a while whether we should include a change-log like that included with ASPECT. We will probably need to continue to change functionality like this, and it would be good to have a place where we can record those changes (and what the user will need to do to their own scripts to get the same output as before. 

What do you think, @tjhei and @ian-r-rose?

tests/test_partitioning.py
```diff
@@ -26,12 +26,12 @@ def test_simple(self):
...
-        pressure = StartP
-        temperature = gt([StartP])[0]
+        pressure = StartP *1.e9
+        temperature = 2000
         (a, b) = burnman.calculate_partition_coefficient(
             pressure, temperature, rel_mol_per, 0.5)
-        self.assertFloatEqual(a, 0.0319126180713)
-        self.assertFloatEqual(b, 0.125395826369)
+        self.assertFloatEqual(a, 0.18453519778)
+        self.assertFloatEqual(b, 0.102938439776)
```
bobmyhill: There's a lot of superfluous code here. Suggest shortening. 
More generally, why the change in values? The bulk composition hasn't changed, and there isn't enough freedom in the system to allow changes in per:bdg ratios.

[The functions in partitioning.py should be replaced with some generalized linear algebra; at the moment it is really clunky. I'll open an issue.]

sannecottaar: Mainly because pressures were in GPa and are now in Pa, but also because I fixed the temperature instead of requestion temperature from the Anderson geotherm.

sannecottaar:
Tests are fixed. 
You want something better than changelog.txt in the main directory?

PR #302, opened 2017-07-04
Add geophysical layer

sannecottaar:
In this pull request I've split up Layer and Planet, for the following reasons:
- We often use a layer kind of setup to evaluate properties in many of our examples. Now these constantly reset the P and T of a Material() and recompute. In Layer, these properties can be cached across the Layer. 
- More importantly, the properties in the layer can be computed with self-consistent pressures. This can be used for a single layer if the pressure at the top and gravity at the bottom are set (the planet as a whole is not self-consistent, but there are applications for this). 
- Geophysical parameters that depend on gradients (e.g. Bullen parameter and brunt-vasala frequency) can be evaluated for layer. 
- Unlike the Layer originally in Planet, adiabats can be used. 
- Both Layer and Planet can evaluate all the properties that are also in a Material (before this was very limited to density, gravity, pressure, and temperature). 

Things that can be implemented going forward:
- I've implemented a modified_adiabat setup for the Layer. This means the temperature of the adiabat can be perturbed, e.g. to add a boundary layer. We could implement functions in geotherm to approximate boundary layers. 
- I can rewrite the seismic output routines for Mineos and AxiSEM to ask for a Layer or Planet, and produce appropriate output file. Similarly I can see Layer being used to produce the input for Aspect. 

You might find that the examples run slower, I think this is mostly due to adiabats being included. 
Let me know your thoughts, and things that can be done better.

### DIFF
--- a/burnman/layer.py
+++ b/burnman/layer.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 import numpy as np
 from scipy.integrate import odeint
 from scipy.integrate import quad
-from scipy.interpolate import UnivariateSpline
+from scipy.interpolate import UnivariateSpline, interp1d
 from burnman import averaging_schemes
 from burnman import constants
 from burnman import geotherm
@@ -23,38 +23,48 @@ import matplotlib.pyplot as plt
 
 class Layer(object):
     """
-    The base class for a planetary layer. The user needs to 
-    1. set_material(), set the material of the layer, either mineral, solid_solution, or composit
-    2. set_temperature_mode(), either predefine, or set an adibat
-    3. set_state(), to set the composition, temperature and pressure of the layer, respecitvely, before properties
-    can be computed. Properties will be returned at the pre-defined depth array, although the evaluate() function
+    The base class for a planetary layer. The user needs to set the following befor properties can be computed
+    1. set_material(), set the material of the layer, e.g. a mineral, solid_solution, or composite
+    2. set_temperature_mode(), either predefine, or set to an adiabat
+    3. set_pressure_mode(), to set the self-consistent pressure (with user-defined option the pressures can be overwritten).
+    To set the self-consistent pressure the pressure at the top and the gravity at the bottom of the layer need to be set.
+    4. make(), computes the self-consistent part of the layer and starts the settings to compute properties within the layer
+    Note that not the entire planet this layer sits in is self-consistent, as the pressure at the top of the layer is a function of the density
+    within the layer (through the gravity). Entire planets can be computed self-consistently with the planet class.
+    Properties will be returned at the pre-defined radius array, although the evaluate() function
     can take a new defined depthlist and values are interpolated between these (sufficient sampling of the layer
     is needed for this to be accurate)
     """
 
-    def __init__(self, name=None, radius_planet=None,
-                 min_depth=None, max_depth=None, n_slices=None, verbose=False):
-
+    def __init__(self, name=None, radii=None, verbose=False):
         self.name = name
-        self.min_depth = min_depth
-        self.max_depth = max_depth
-        self.thickness = self.max_depth - self.min_depth
-        self.n_slices = n_slices
-        self.depths = np.linspace(min_depth, max_depth, n_slices)
-        self.radius_planet = radius_planet
-        self.radii = self.radius_planet - self.depths
+        self.radii = np.sort(radii)
         self.outer_radius = max(self.radii)
         self.inner_radius = min(self.radii)
+        self.thickness = self.outer_radius - self.inner_radius
+        self.n_slices = len(self.radii)
         self.verbose = verbose
         self._cached = {}
         self._pressures = None
         self._temperatures = None
         self.sublayers = None
-    
+        self.material = None
+        self.pressure_mode = None
+        self.temperature_mode = None
+ 
+ 
+    def __str__(self):
+        """
+        Prints details of the Layer
+        """
+        writing  =  ' Layer ' + self.name + ' made out of ' + str(self.material.name) + ' with ' \
+                        + str(self.temperature_mode) + ' temperatures  and ' + str(self.pressure_mode) + ' pressures \n'
+        return writing
+            
+            
     def reset(self):
         """
         Resets all cached material properties.
-
         It is typically not required for the user to call this function.
         """
         self._cached = {}
@@ -79,9 +89,9 @@ class Layer(object):
         Parameters
         ----------
         temperature_mode : string
-        This can be set to 'user-defined','isothermal','adiabatic', or 'modified-adiabat'
+        This can be set to 'user-defined', 'adiabatic', or 'modified-adiabat'
         temperatures : array of float
-        The desired fixed temperatures in [K]. Should have same length as defined depths in layer.
+        The desired fixed temperatures in [K]. Should have same length as defined radii in layer.
         temperature_top : float
         Temperature at the top for an adiabat
         
@@ -93,8 +103,7 @@ class Layer(object):
             e.g. allows to implement boundary layers
         """
         self.reset()
-        assert(temperature_mode == 'user-defined' or temperature_mode ==
-               'isentropic', temperature_mode ==
+        assert(temperature_mode == 'user-defined'  or temperature_mode ==
                'adiabat' or temperature_mode == 'modified-adiabat')
 
         self.temperature_mode = temperature_mode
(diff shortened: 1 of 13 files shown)


sannecottaar:
Hmm, one of the tests fails. A lot of them do on my computer, as I always find slightly different numbers. Can someone else fix this?

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+from .mineral import Mineral
+from .composite import Composite
+from .seismic import Seismic1DModel
+
+import matplotlib.pyplot as plt
+
+
+class Layer(object):
+    """
+    A planetary layer class
```
bobmyhill: This is going to be the main summary in the autogenerated docs. It would be good to describe in more detail what this class is, and what it does.

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+    """
+
+    def __init__(self, name=None, radius_planet=None,
```
bobmyhill: radius_planet? Why is this needed in layer?

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+                 min_depth=None, max_depth=None, n_slices=None, verbose=False):
+
+        self.name = name
+        self.min_depth = min_depth
+        self.max_depth = max_depth
+        self.thickness = self.max_depth - self.min_depth
+        self.n_slices = n_slices
+        self.depths = np.linspace(min_depth, max_depth, n_slices)
```
bobmyhill: On balance, I think I'd prefer to see a single "depth slices" in the initialiser which takes an array of slice depths. That way, the user can provide non-uniform slices, and this also meshes more nicely with the user_defined temperature/pressure inputs below.

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        The desired fixed temperatures in [K]. Should have same length as defined depths in layer.
+        temperature_top : float
+        Temperature at the top for an adiabat
+        
+        Note
+        ---------
+        'user_defined' = fixes the temperature with the profile input by the users
+        'adiabat' = self-consistently computes the adiabat when setting the state of the layer
+        'modified_adiabat' = adds the user input array to the adiabat, 
+            e.g. allows to implement boundary layers
```
bobmyhill: I like this. We should also consider implementing the McKenzie (2005) boundary layer model (or a more modern alternative). I've been using it for Mars, and I think it would be handy for others as well.

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+            self.usertemperatures = temperatures
+        else:
+            self.usertemperatures = np.zeros_like(self.depths)
+
+        if temperature_mode == 'adiabat' or temperature_mode == 'modified_adiabat':
+            self.temperature_top = temperature_top
+        else:
+            self.temperature_top = None
+
+    def set_state( self, pressure_mode='selfconsistent', pressures=None, pressure_top=None,
```
bobmyhill: I would prefer 'self-consistent'. Similarly, I think I prefer 'user-defined' for the temperature_mode and pressure_mode names.

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        Used by set_state()
+        """
+        if self.temperature_mode == 'adiabat' or self.temperature_mode == 'modified_adiabat':
+            adiabat = geotherm.adiabatic(
+                pressures, temperature_top, self.material)
+        else:
+            adiabat = np.zeros_like(self.depths)
+        return adiabat + self.usertemperatures
+
+    def _evaluate_eos(self, pressures, temperatures, gravity_bottom, pressure_top):
```
bobmyhill: I think the choice of gravity_bottom and pressure_top needs some documentation, both here and briefly in the Class blurb.

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        density = self.density[::-1]
+        rhofunc = UnivariateSpline(radii, density)
+        moment = np.abs(quad(lambda r: 8.0 / 3.0 * np.pi * rhofunc(r)
+                             * r * r * r * r, radii[0], radii[-1])[0])
+        return moment
+
+    @property
+    def gravity(self):
+        """
+        Returns gravity of the layer [m s^(-2)]
```
bobmyhill: gravity -> gravity profile?

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        """
+        return self._compute_gravity(self.density, self.gravity_bottom)
+
+    @property
+    def bullen(self):
+        """
+        Returns the Bullen parameter
```
bobmyhill: Definition needed

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        dkappadP = np.gradient(kappa, edge_order=2) / \
+            np.gradient(self.pressure, edge_order=2)
+        dphidz = np.gradient(phi,
+                             edge_order=2) / np.gradient(self.depths,
+                                                         edge_order=2) / self.gravity
+        bullen = dkappadP - dphidz
+        return bullen
+
+    @property
+    def brunt_vasala(self):
```
bobmyhill: Description definitely needed!

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        kappa = self.bulk_sound_velocity * self.bulk_sound_velocity * self.density
+        brunt_vasala = self.density * self.gravity * \
+            self.gravity * (self.bullen - 1.) / kappa
+        return brunt_vasala
+
+    @property
+    def pressure(self):
+        """
+        Returns current pressure that was set with :func:`~burnman.material.Material.set_state`.
```
bobmyhill: Is this an array of pressures for each of the depths? This isn't clear.

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        pressure : array of floats
+            Pressure in [Pa].
+        """
+        return self._pressures
+
+    @property
+    def temperature(self):
+        """
+        Returns current temperature that was set with :func:`~burnman.material.Material.set_state`.
+
```
bobmyhill: Same here...

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        -------
+        temperature : array of floats
+            Temperature in [K].
+        """
+        return self._temperatures
+
+    @material_property
+    def internal_energy(self):
+        """
+        Returns the internal energy of the layer.
```
bobmyhill: Do you mean the internal energy at each pressure-temperature point? Again, this isn't clear.

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        internal_energy : array of floats
+            The internal energy in [J].
+        """
+        return np.array(
+            [self.sublayers[i].internal_energy for i in range(len(self.sublayers))])
+
+    @material_property
+    def molar_gibbs(self):
+        """
+        Returns the Gibbs free energy of the layer.
```
bobmyhill: And again...

burnman/layer.py
```diff
@@ -0,0 +1,797 @@
...
+        molar_gibbs : array of floats
+            Gibbs free energy in [J].
+        """
+        return np.array(
+            [self.sublayers[i].molar_gibbs for i in range(len(self.sublayers))])
+
+    @material_property
+    def molar_helmholtz(self):
+        """
+        Returns the Helmholtz free energy of the layer.
```
bobmyhill: And so on (I won't comment all of these)

burnman/planet.py
```diff
@@ -7,206 +7,796 @@
...
 from scipy.interpolate import UnivariateSpline
-from burnman import averaging_schemes
 from burnman import constants
+from .material import Material, material_property
 
 
 class Planet(object):
     """
-    A planet class that find a self-consistent planet.
+    A planet class that finds a self-consistent planet.
```
bobmyhill: This needs more description.

burnman/planet.py
```diff
@@ -7,206 +7,796 @@
...
+        
+            if depthlist is not None:
+                oneprop = np.interp(depthlist, self.depths,oneprop)
+            if all is None:
+                all = np.array(oneprop)
+            else:
+                np.append(all, oneprop, axis=0)
+        return all
+
+    def set_state( self, pressure_mode='selfconsistent', pressures=None, pressure_top=0.,
```
bobmyhill: 'self-consistent'

burnman/planet.py
```diff
@@ -7,206 +7,796 @@
...
+            gravity_bottom=0., n_max_iterations=50):
+        """
+        Sets the pressure of the planet by user-defined values are in a self-consistent fashion.
+        pressure_mode is 'user_defined' or 'selfconsistent'.
```
bobmyhill: 'user-defined'

burnman/planet.py
```diff
@@ -7,206 +7,796 @@
...
         for layer in self.layers:
-            radii = self.radial_slices[layer.n_start: layer.n_end]
-            density = self.densities[layer.n_start: layer.n_end]
-            rhofunc = UnivariateSpline(radii, density)
-            moment += quad(lambda r : 8.0/3.0*np.pi*rhofunc(r)*r*r*r*r,
-                           radii[0], radii[-1])[0]
+            moment += layer.moment_of_inertia
         return moment
+
+    @property
```
bobmyhill: From here, my comments are very similar to those in the other file...

burnman/seismic.py
```diff
@@ -427,7 +473,8 @@ class STW105(SeismicTable):
 
     def __init__(self):
         SeismicTable.__init__(self)
-        table = tools.read_table("input_seismic/STW105.txt") # radius, pressure, density, v_p, v_s
+        # radius, pressure, density, v_p, v_s
+        table = tools.read_table("input_seismic/STW105.txt")
```
bobmyhill: I'd quite like to see read_table be deprecated by np.loadtxt or np.genfromtxt - I wrote it before I knew much python, and I think it would be better to use standard (and more powerful) tools where possible.

burnman/seismic.py
```diff
@@ -438,11 +485,23 @@ def __init__(self):
         self.table_vph = table[:, 6][::-1]
         self.table_vsh = table[:, 7][::-1]
 
-        self.table_depth = self.earth_radius-self.table_radius
+        self.table_depth = self.earth_radius - self.table_radius
 
         # Voigt averages for Vs and Vp
-        self.table_vs = np.sqrt((2.*self.table_vsv*self.table_vsv+self.table_vsh*self.table_vsh)/3.)
-        self.table_vp = np.sqrt((self.table_vpv*self.table_vpv+4.*self.table_vph*self.table_vph)/5.)
+        self.table_vs = np.sqrt(
```
bobmyhill: I don't think so many line breaks make this any more readable.

examples/example_build_planet.py
```diff
@@ -3,35 +3,6 @@
...
-to be able to make a model of the planet's interior that is consistent with those
-measurements.
-
-However, there is a difficulty with this.  In order to know the density of the planetary
-material, we need to know the pressure and temperature.  In order to know the pressure,
-we need to know the gravity profile.  And in order to the the gravity profile, we need
-to know the density.  This is a nonlinear problem which requires us to iterate to find
-a self-consistent solution.
-
-Here we show an example that does this, using the planet Mercury as motivation.
```
bobmyhill: I think this was a good docstring for the manual. Could you update it rather than removing it?

examples/example_layer.py
```diff
@@ -0,0 +1,140 @@
...
+        ['pressure', 'gravity', 'density', 'v_p', 'v_s', 'bullen'], depths)
+
+    # Here we define the lower mantle as a Layer(). The layer needs various
+    # parameters to set a depth array and radius array.
+    lower_mantle = burnman.Layer( name='Lower Mantle', radius_planet=6371.e3,
+        min_depth=750.e3, max_depth=2800.e3,  n_slices=20)
+    # Here we set the composition of the layer as the above defined 'rock'.
+    lower_mantle.set_material(rock)
+
+    # Now we set the temperature mode of the layer. , and a
```
bobmyhill: . ,

examples/example_layer.py
```diff
@@ -0,0 +1,140 @@
...
+    # self-consistent pressure. The pressure at the top of the layer and
+    # gravity at the bottom of the layer are given by the PREM.
+    
+    # Here we use an adiabatic temperature and set the temperature at the top of the layer
+    lower_mantle.set_temperature_mode(temperature_mode='adiabat', temperature_top=1900.)
```
bobmyhill: Nice :)

examples/example_layer.py
```diff
@@ -0,0 +1,140 @@
...
+                           
+    lower_mantle.set_state(pressure_mode='selfconsistent',
+                               pressure_top=pressure[0], gravity_bottom=gravity[-1])
+
+       
+    # All the work is done, now we can plot various properties!
+    # First, we plot the s-wave speed verses the PREM s-wave speed
+    plt.figure(figsize = (10,6))
+    plt.subplot(1, 2, 1)
+    plt.plot(lower_mantle.pressure / 1.e9, lower_mantle.v_s / 1.e3, color='b', linestyle='-',
```
bobmyhill: I feel that these properties would be better pluralised (saving  the singular for individual "evaluate" calls), but if you have strong arguments against doing that, then I'm all ears.

misc/ref/example_build_planet.py.out
```diff
@@ -1,17 +1,22 @@
...
+Iteration 4  maximum core pressure error between iterations: 6.843174e-02
+Iteration 5  maximum core pressure error between iterations: 2.556444e-02
+Iteration 6  maximum core pressure error between iterations: 9.184659e-03
+Iteration 7  maximum core pressure error between iterations: 3.248780e-03
+Iteration 8  maximum core pressure error between iterations: 1.142705e-03
+Iteration 9  maximum core pressure error between iterations: 4.011601e-04
+Iteration 10  maximum core pressure error between iterations: 1.407470e-04
+Iteration 11  maximum core pressure error between iterations: 4.937207e-05
+Iteration 12  maximum core pressure error between iterations: 1.731808e-05
+Iteration 13  maximum core pressure error between iterations: 6.074510e-06
```
bobmyhill: error -> delta?

examples/example_build_planet.py
```diff
@@ -81,86 +51,126 @@
...
+    # Separately the composition and the temperature_mode need to set.
+    radius_planet = 6371.e3
+    # inner_core
+    inner_core = burnman.Layer("inner core", radius_planet=radius_planet,
+        max_depth=6371.e3, min_depth=5151.e3, n_slices=10)
+    inner_core.set_material(burnman.minerals.other.Fe_Dewaele())
+    
+    # The minerals that make up our core do not currently implement the thermal equation of state, so we will define the temperature with the Brown & Shankland geotherm.
+    inner_core.set_temperature_mode( 'user_defined',
+        burnman.geotherm.brown_shankland(inner_core.depths))
```
CaymanUnterborn: Should this just be changed to 300K or NaN's? If we're not calculating Temperature it doesn't make sense to say "Oh it's Brown and Shankland." That could confuse users, particularly ones who aren't expressly aware that these EoS are not T dependent.

examples/example_build_planet.py
```diff
@@ -81,86 +51,126 @@
...
+
+
+
+    # Now we calculate the planet.
+    Plan = burnman.Planet('earth_like',
+                          [inner_core, outer_core, lower_mantle, upper_mantle], verbose=True)
+    # Here we compute its state. Go BurnMan Go!
+    # (If we were to change composition of one of the layers, we would have to
+    # recompute the state)
+    Plan.set_state()
```
CaymanUnterborn: I really like that this is a set_state now. Flows well with the rest of the functionality of BurnMan

examples/example_build_planet.py
```diff
@@ -81,86 +51,126 @@
...
+    outer_core.set_temperature_mode('user_defined',
+        burnman.geotherm.brown_shankland(outer_core.depths))
+
+    # Next the Mantle.
+    lower_mantle = burnman.Layer( "lower_mantle", radius_planet=radius_planet,
+        max_depth=2891.e3, min_depth=660.e3, n_slices=10)
+    lower_mantle.set_material(burnman.minerals.SLB_2011.mg_bridgmanite())
+    lower_mantle.set_temperature_mode('adiabat')
+    upper_mantle = burnman.Layer( "upper_mantle", radius_planet=radius_planet,
+        max_depth=660e3, min_depth=0., n_slices=10)
```
CaymanUnterborn: I'm confused why a Layer needs to know the radius of the planet. Wouldn't it just be the sum of the min_depth's sent to the Planet class yield the same result?

burnman/layer.py
sannecottaar: To compute both radius and depth profiles, i.e. needed for gravity and mass computations.

examples/example_build_planet.py
sannecottaar: Layer needs it to be used independently from the planet class (to have both depth and radius arrays). I have a plan to build a routine that can build the planet without defining the individual layers first, maybe I will include this now.

examples/example_layer.py
sannecottaar: the argument against would be to have a consistent library of property names (and aliases) across material, layer and planet.

burnman/layer.py
bobmyhill: Hmmm, I don't think the user should be able to know the depth from the surface of a planet when they've just only declared a layer... radius is a much more natural choice.

For a layer, we only need to know the top_radius, bottom_radius, and pressure and gravity at either the top or the bottom or one of each (which could conceivably be user choice, though that might be messy). Mass and gravity computations don't require that we know anything about the planet away from the boundaries (Gauss' law).

examples/example_build_planet.py
sannecottaar: Clearly I'm not aware either ;) Do we have any parameters in Burnman for a thermal EoS for iron?

bobmyhill: I wish. There's the AA1994 model that's been a PR for ages. I should make a decision on this at some point.

bobmyhill: Oh, wait ... inner core? No, but it wouldn't be hard to add.

burnman/layer.py
sannecottaar: Yes, that might be good.

burnman/layer.py
CaymanUnterborn: That would be awesome Bob!

burnman/layer.py
sannecottaar: I think we also had depth or radius discussions in December. It does seem like planetary scientists use radius. However, so far in BurnMan, we've used depth a lot. I guess mostly our seismic models are defined by depth.

bobmyhill: We definitely discussed this, yes. I don't know what the conclusions were. I don't think radius_planet should go in the layer init. We also risk ambiguity: is this the depth from the top of the layer, or depth from the surface? For a planet it's a different story, but I don't think many users will find it such a chore to write `depths = outer_radius - radii`

sannecottaar:
The old planet class was definitely radius focused, so probably that was
the outcome of the discussion. Just working on this, I found Burnman very
depth-focused, which is what we agreed on before seismic models. It
probably also comes out of everything being computed with increasing
pressure (i.e. most computations, besides gravity, would be with decreasing
radius).  I preferred to keep it consistent. But not having to know depth
for a layer, would simplify to have everything radii. Might be worth
talking about.

tjhei:
> I can rewrite the seismic output routines for Mineos and AxiSEM to ask for a Layer or Planet, and produce appropriate output file. Similarly I can see Layer being used to produce the input for Aspect.

I would like to keep the concept of a layer that computes stuff separate from a data table to import/export data from. Otherwise one would need to construct a layer and somehow stuff data into it when trying to export manual data.

burnman/seismic.py
sannecottaar: Yes, but I've ignored this for now.

sannecottaar:
I think, I've covered most of your comments. I left all the properties in for now, as I prefer to call them directly, but they could be taken out. Timo has other ideas about a generalized data table.

Unfortunately the test fails. As usual my computer give quite significant different values...

I'll be out of town for the next two weeks, so I won't be changing anything. Feel free to merge and make your own changes. I meant for this to be a proposed design change, and not a finalized version. We can have this out in a new release when we have a paper (so in lots of time).

sannecottaar:
Would anybody have time to merge this and fix the test (I cannot fix this from my computer)? This might not be in the final stage yet, but I think it is good to merge so everybody can make improvements. I would like to develop further on including thermal boundary layers.

bobmyhill:
@tjhei Could you run tests? I think this is ready to merge.

sannecottaar:
The error that remains here seems to be a numpy version issue:
TypeError: gradient() got an unexpected keyword argument 'edge_order'
This was implemented in numpy 1.9.1. Do we need to be compatible with older versions and fix this?

bobmyhill:
@sannecottaar 

At the moment, yes, we need to be backwards compatible. However, all of the numpy gradient and image routines are really useful, and I think we should update our dependencies (@tjhei, can that be soon?).

For the time being, how about a try:except:, with the except simply removing the edge_order term? If you don't compare the first and last element of the output in your tests, the output should be identical.

One last thing; you can take the gradient with an extra term for the x-spacing (i.e. np.gradient(f, x)). Might make that bit of code a bit tidier.

burnman/planet.py
```diff
@@ -7,206 +7,834 @@
...
         """
         moment = 0.0
         for layer in self.layers:
-            radii = self.radial_slices[layer.n_start: layer.n_end]
-            density = self.densities[layer.n_start: layer.n_end]
-            rhofunc = UnivariateSpline(radii, density)
-            moment += quad(lambda r : 8.0/3.0*np.pi*rhofunc(r)*r*r*r*r,
-                           radii[0], radii[-1])[0]
+            moment += layer.moment_of_inertia
         return moment
```
bobmyhill: This function and the ones for gravity and mass above are a bit cumbersome. How about:
`return np.sum([layer.moment_of_inertia for layer in self.layers])`

(There's actually a lot more in this PR that could be made more readable with list comprehensions, but the ones starting with empty lists or sums starting from 0.0 caught my eye.)

burnman/layer.py
```diff
@@ -0,0 +1,899 @@
...
+
+    def set_material(self, material):
+        """
+        Set the material of a Layer with a Material
+        """
+        assert(isinstance(material, Material))
+        self.material = material
+        self.reset()
+
+    def set_temperature_mode(self, temperature_mode='adiabat',
```
bobmyhill: There needs to be some check that temperature_top has been provided for a single layer adiabatic planet. Also a test would be good.

burnman/layer.py
```diff
@@ -0,0 +1,899 @@
...
+
+
+        if self.pressure_mode == 'user-defined':
+            self.temperatures = self._evaluate_temperature(
+                self._pressures, self.temperature_top)
+
+        if self.pressure_mode == 'self-consistent':
+            new_press = self.pressure_top + \
+                (-self.radii + max(self.radii)) * \
+                2.e5  # initial pressure curve guess
```
bobmyhill: 2.e5 is a really big number. Is it possible to make a more reasonable guess? Otherwise lots of low pressure EoSes are going to fail here.

burnman/planet.py
```diff
@@ -7,206 +7,834 @@
...
+
+
+        if self.pressure_mode == 'user-defined':
+            self._temperatures = self._evaluate_temperature(self._pressures)
+
+        if self.pressure_mode == 'self-consistent':
+
+            new_press = self.pressure_top + \
+                (-self.radii + max(self.radii)) * \
+                2.e5  # initial pressure curve guess
```
bobmyhill: 2.e5 is a really big number. Is it possible to make a more reasonable guess? Otherwise lots of low pressure EoSes are going to fail here.

bobmyhill:
@sannecottaar have you tested evaluate with >1 property? Can you add this functionality to the example? [Only the last property is being returned]

PR #304, opened 2017-07-07
Fix DKS equations of state tests

bobmyhill:


bobmyhill:
Just a few output changes so that the tester passes. I'll merge this now.

Issue #307, opened 2017-07-18
add energies to static equations of state


bobmyhill:
It would be good to add both the internal/helmholtz energy and gibbs/enthalpy to static equations of state, as these are both useful in their own right and also key components of thermal equations of state. 

`P = -dU/dV | S = -dA/dV | T`
`V = dG/dP | T = dH/dP | S`

At 0 K, S=0 J/K/mol. 
The integration only needs to be done for one of the four potentials; Legendre transformations will take care of the others. Which integration is easiest (VdP or PdV) depends on whether the EoS is explicit in V or P.

PR #308, opened 2017-07-18
added energies to static equations of state

bobmyhill:
addresses Issue #307


tjhei:
Looks reasonable but the code is not tested or even executed in an example, right?

bobmyhill:
Excellent point. Tests now added.

PR #309, opened 2017-07-31
Added flexibility and NaN processing to reading of PerpleX tables

bobmyhill:
The latest version of PerpleX has changed the ordering of P and T. This PR adds a check to ensure that pressure and temperature are read correctly. It also attempts to replace NaNs from the PerpleX tables via interpolation using pandas (if it is installed).

### DIFF
--- a/burnman/perplex.py
+++ b/burnman/perplex.py
@@ -12,7 +12,7 @@ from os import rename
 
 
 import numpy as np
-from scipy.interpolate import interp2d
+from scipy.interpolate import interp2d, griddata
 
 from .material import Material, material_property
 from . import eos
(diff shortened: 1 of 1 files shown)


burnman/perplex.py
```diff
@@ -105,26 +105,33 @@ def _read_2D_perplex_file(self, filename):
...
-        Pint = float(lines[5][0])*1.e5
-        nP = int(lines[6][0])
+        bounds = [(float(lines[4][0]), float(lines[5][0]), int(lines[6][0])),
+                  (float(lines[8][0]), float(lines[9][0]), int(lines[10][0]))]
+
+        if lines[3][0] == 'P(bar)':
+            Pmin, Pint, nP = bounds[0]
+            Tmin, Tint, nT = bounds[1]
+        else:
+            Pmin, Pint, nP = bounds[1]
```
tjhei: can you assert here that this is indeed the other ordering?

tjhei:
what does pandas do with the NaN values and what does scipy do?

bobmyhill:
if scipy encounters nan when trying to interpolate, it returns zero. pandas assumes that a patch of nans  represents a gap in data coverage (or an error), and tries to fill that gap by interpolation.

bobmyhill:
I've removed pandas in favour of a scipy workaround. It'll be a bit slower, but probably preferable to requiring pandas. That said, we may soon want to add pandas as a dependency; it'll make quite a lot of data processing easier.

PR #310, opened 2017-08-04
Energy calculations for the reciprocal K' EoS

bobmyhill:
If there's an analytical solution, it's messy...

### DIFF
--- a/burnman/eos/reciprocal_kprime.py
+++ b/burnman/eos/reciprocal_kprime.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 
 
 import scipy.optimize as opt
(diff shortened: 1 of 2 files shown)


bobmyhill:
Turns out there is an analytical expression, but it requires the gammainc function with a<0, which is not supplied by scipy.special. It's possible to use the mpmath implementation, but this requires another package. Thoughts?

tjhei:
I would think it is easier to just use the numerical integration.

bobmyhill:
I think this is the best solution. It avoids quad integration while only using scipy functions. It's also quicker than using mpmath. a is a small negative number, so the recurrence relation loop is pretty speedy.

PR #311, opened 2018-01-18
Fix solution potentials

bobmyhill:
Well, this is embarrassing. It seems no-one ever tried to calculate Helmholtz or internal energy for a solid solution, and both were implemented incorrectly. This PR fixes the errors and adds the appropriate test.

minerals and materials have always been implemented correctly.


bobmyhill:
Sure, I thought it was important in this case!! Ok, I'll merge now, as this is a real bug.

PR #312, opened 2018-02-06
Enabled combined minerals to be used in solid solutions

bobmyhill:
The CombinedMineral class was derived in such a way that when included in a solid solution, it would lose its method. This PR makes the class more like the other minerals, with a method set using the equation_of_state param and burnman/eos/helper.py.

### DIFF
--- a/burnman/combinedmineral.py
+++ b/burnman/combinedmineral.py
@@ -35,14 +35,15 @@ class CombinedMineral(Mineral):
     This class is available as :class:`burnman.CombinedMineral`.
     """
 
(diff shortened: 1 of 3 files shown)


sannecottaar:
Looks fine. Not sure I fully understand the class though. Where do the 'free energy adjustments' get used? How our they linear in pressure and temperature? Defining the units for the three adjustments might help me there.

bobmyhill:
Ok, I've added the free energy adjustment units to the blurb. CombinedMinerals like this are used by Tim Holland all the time; it's actually pretty amazing how close we can get to the properties of new phases just by adding others together.

PR #315, opened 2018-03-21
Add damped newton solver, tidy solution model processing

bobmyhill:
The PR does three things in preparation for an equilibrium gibbs calculator:
- Add a derivative-based damped affine invariant newton solver (Deuflhard, 1974;1975;2004) with optional constraints. I have added several standard tests for the solver.
- Neaten up the solution model chemical processing. This is mostly cosmetic, although I have added a new return attributes (site_names) and function (solution_bounds), which will make gibbs calculations more straightforward.
- Add a __str__ function to burnman.composite.Composite for pretty printing purposes.


tests/test_solvers.py
```diff
@@ -0,0 +1,112 @@
+import unittest
```
tjhei: is this executed or do we need to list it in ``tests.py``?

burnman/composite.py
```diff
@@ -72,6 +73,21 @@ def __init__(self, phases, fractions=None, fraction_type='molar'):
             self.molar_fractions = None
 
         self.set_averaging_scheme('VoigtReussHill')
+        self.name=name
+        
+    def __str__(self):
```
tjhei: This is unrelated to the PR. I assume you want it for the tests? We have ``to_string()`` in Material, Composite, etc.  which is not very pythonic but and this is very similar to ``debug_print`` (did you see this function?). The ``self.name`` should probably appear in ``to_string()`` and/or ``debug_print`` if you insist on keeping things this way.

burnman/nonlinear_solvers.py
```diff
@@ -0,0 +1,212 @@
...
+    with Jacobian J(x) and constraints C(x), using the 
+    damped affine invariant modification to Newton's method 
+    (Deuflhard, 1974;1975;2004), 
+    Here we follow the algorithm as described in Nowak and Weimann (1991):
+    [Technical Report TR-91-10, Algorithm B]
+
+    The iteration continues until three conditions are satisfied:
+    all(np.abs(dx (simplified newton step) < tol))
+    all(np.abs(dx (full Newton step) < sqrt(10*tol))) [avoiding pathological behaviour] and
+    lambda = lambda_bounds(dx, x)[1] (lambda = 1 for a full Newton step).
```
tjhei: are you missing a comparison here (``lambda<0`` or something?)?

burnman/nonlinear_solvers.py
```diff
@@ -0,0 +1,212 @@
...
+
+    Inequality constraints are provided by the function C(x), which returns a
```
tjhei: there is no ``C`` argument

burnman/nonlinear_solvers.py
```diff
@@ -0,0 +1,212 @@
...
+        Maximum number of iterations for solver.
+    lambda_bounds: function
+        Returns a tuple of floats (1.e-8, 1.) corresponding
+        to the minimum and maximum allowed fractions of the
+        full newton step (dx). A function of dx and x.
+    constraints : function
+        Returns the LHS of the inequality constraints(x)
+        as a 1D numpy array. The constraints are satisfied if
+        all the elements of the array are less than or equal to
+        zero.
```
tjhei: "A function of x"

burnman/composite.py
bobmyhill: Yes, I should probably have put this in a separate PR. 
I think it might be worthwhile replacing `to_string()` with `__str__()` in all classes. That way, when anyone wants to see key information about an object, they can just type `print(object)`, rather than `print(object.to_string())`. What do you think?

bobmyhill:
matplotlib was also giving me a bunch of deprecation warnings, so I modernised the plotting commands in a couple of examples.

burnman/composite.py
bobmyhill: For the time being, I've just added self.name to to_string and debug_print.

PR #317, opened 2018-03-22
Added gibbs, entropy and volume hessian calculations to solid solutions

bobmyhill:
For stability in gibbs minimization / equilibrium calculation schemes, it is useful to have analytical expressions for the compositional gibbs hessian (d2G/dxidxj), which is equal to the compositional gradient of the chemical potentials of the endmembers (dmui/dxj).

It is also useful to have the first derivatives of the gibbs hessian with respect to temperature and pressure - the entropy and volume Hessians.

This PR implements these expressions for the ideal, symmetric and asymmetric solution models.
This branch is derived from that in PR #316

### DIFF
--- a/burnman/composite.py
+++ b/burnman/composite.py
@@ -76,6 +76,7 @@ class Composite(Material):
     def set_fractions(self, fractions, fraction_type='molar'):
         """
         Change the fractions of the phases of this Composite.
+        Resets cached properties
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


sannecottaar:
As far as I can tell, this looks good.

bobmyhill:
Great, thanks @sannecottaar! If #316 also looks good, then I think they're both ready to go :)

PR #318, opened 2018-03-24
Added composition functions to Composite class

bobmyhill:
This PR is now largely obsolete because of other smaller PRs (#345, #346, #347, #348).

The only new thing in here now is the composition functions in the composite class, which I don't like all that much.


bobmyhill:
Requires sympy

bobmyhill:
This PR is now largely obsolete because of other smaller PRs (#345, #346, #347, #348).

The only new thing in here now is the composition functions in the composite class, which I don't like all that much.

PR #320, opened 2018-04-06
Implement bound-walking for the non-linear solver

bobmyhill:
Building on PR #315, this PR implements bounds-walking for the affine invariant newton solver. 

Instead of terminating if the Newton descent path leads outside the feasibility region, the solver now uses the method of Lagrangian multipliers to 
minimize `||F(x_i) + J(x_i) dx||`
subject to `C(x_i + dx) = 0`

where `C(x_i + dx)` are the constraints for the solver to walk along. 

I've added some more tests. Because this method only works for constraints which are linear in the parameters `x`, I've removed the ability to pass arbitrary constraint functions to the solver.

### DIFF
--- a/burnman/nonlinear_solvers.py
+++ b/burnman/nonlinear_solvers.py
@@ -2,25 +2,53 @@ import numpy as np
 from scipy.linalg import lu_factor, lu_solve
 from collections import namedtuple
 
-def solve_constraint_lagrangian(x_n, J, c_newton, c_A):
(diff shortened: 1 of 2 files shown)


sannecottaar:
Sorry, did you want this merged instead of #315?

bobmyhill:
Nah, it was better to merge #315 first. I've rebased to master, so it should be easier to review this now. Thanks!

PR #321, opened 2018-04-11
Promote DOI

katrinleinweber:
Hello, and thanks for publishing this :-)

Because the citation info is currently only on the website, I thought it might help users to also find it here.
Or should [10.1002/2013GC005122](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1002/2013GC005122) be cited primarily?

Cheers!


bobmyhill:
Thanks @katrinleinweber!

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #323, opened 2018-04-13
WIP: Equilibrate

bobmyhill:
Here's the equilibrium relations branch that I've been working on. The functions in equilibrate are still not fully commented, and I need to add some tests, but the front-end functionality is working well for the problems I've tried.

This PR depends on sympy, which is not currently installed on the tester.

Some examples of things we can now do are given in `example_equilibrate.py`. Any equilibration problem can be formulated and passed to the solver in 4 lines:

1. bulk composition definition (as a dictionary of elements)
2. assemblage definition (as a burnman.Composite)
3. equality constraints (as a list of two tuples)
4. call to the solver: returns the solution(s) and a lot of other useful information in two namedtuples.

For the next PR, I would like to add automatic 1D/2D phase diagram construction (essentially applying Schreinemakers' rules). However, that is definitely beyond the scope of this PR.

Some things still to do:
- [ ] Use sparse methods for Jacobian LU decomposition: https://docs.scipy.org/doc/scipy/reference/generated/generated/scipy.sparse.linalg.splu.html
- [ ] Remove starting guess calculations, which are slow and might even inhibit convergence


burnman/processchemistry.py
```diff
@@ -59,7 +59,7 @@ def dictionarize_formula(formula):
     return f
 
 def sum_formulae(formulae, amounts=None):
-    if amounts == None:
+    if amounts is None:
```
sannecottaar: I cannot comment on non-changes.... but zip() in line 143 needs to a list() around it for python3.

burnman/equilibrate.py
```diff
@@ -0,0 +1,775 @@
...
+        cons_fun = lambda x:  R.dot(baseline_endmember_amounts + reaction_vectors.dot(x)) # >= 0
+        cons = ({'type': 'ineq', 'fun': cons_fun})
+
+        sol = minimize(func_min_cdist, [0.]*reaction_vectors.shape[1],
+                       args=(guess_proportions, baseline_endmember_amounts,
+                             reaction_vectors, prm.indices),
+                       method='COBYLA', constraints=cons, tol=1.e-16)
+
+        #assert(sol.status==1)
+        assert(cons_fun(sol.x) > -eps)
```
sannecottaar: python3 doesn't like comparing a list to a float. Is it enough to check that the minimum of the cons_fun output is > -eps?

burnman/equilibrate.py
```diff
@@ -0,0 +1,775 @@
...
+                        updated_params = True
+                if not updated_params:
+                    prm.baseline_endmember_amounts = calculate_baseline_endmember_amounts(assemblage,
+                                                                                          cs,
+                                                                                          prm)
+                    prm.initial_parameters = get_parameters_from_state_and_endmember_amounts(initial_state, assemblage, prm)
+
+    # Finally, make dimensions of sol_list equal the input dimensions 
+    if len(sol_list[0]) == 1:
+        sol_list = list(zip(*sol_list)[0])
```
sannecottaar: closing bracket for list() needs to be before [0]

examples/example_equilibrate.py
```diff
@@ -0,0 +1,373 @@
...
+import burnman
+from burnman.minerals import HP_2011_ds62, SLB_2011, JH_2015
+from burnman import equilibrate
+
+ordering = False # This example plots the state of order of the Jennings and Holland orthopyroxene in the simple en-fs binary at 1 bar.
+aluminosilicates = False # The example plots the andalusite-sillimanite-kyanite phase diagram
+gt_solvus = False # This example demonstrates the shape of the pyrope-grossular solvus
+lower_mantle = True # This example calculates temperatures and assemblage properties along an isentrope in the lower mantle
+upper_mantle = False # This example produces a 2D grid of the ol-opx-gt field
+olivine_polymorphs = True # DOESN'T SOLVE IF LOWER MANTLE OR UPPER MANTLE ARE RUN BEFOREHAND. NEED TO ROOT OUT THIS WEIRD BUG! This example produces a P-T pseudosection for a fo90 composition
```
sannecottaar: Is this bug solved? At least it was no problem for me.

bobmyhill: Yes, I fixed this yesterday. Repeated function calls in python are peculiar. I added a note at the very top of equilibrate to remind me not to reverse the change in the future.

Issue #331, opened 2018-09-10
Add sympy to tester


bobmyhill:
sympy is a really valuable package for a lot of thermodynamics. It would be really good to have this installed on the tester

PR #334, opened 2018-10-01
add Jenkins integration

tjhei:
- new testing infrastructure based on https://jenkins.tjhei.info/
- update test script to fail if error is encountered
- add sympy support (and a test for it)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@
 //
 // docker run --rm -v "$(pwd):/src" tjhei/burnman:v4 /bin/bash -c "cd /src && ./test.sh"
 //
-// This requires and installation of docker ( https://docs.docker.com/get-started/ )
+// This requires installation of docker ( https://docs.docker.com/get-started/ )
 // and ~ 2GB of images will be downloaded.
 
 pipeline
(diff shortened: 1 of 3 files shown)


tjhei:
closes #331

Jenkinsfile
```diff
@@ -0,0 +1,145 @@
...
+  stages
+  {
+
+    stage("check")
+    {
+      agent
+      {
+        docker
+        {
+          image 'dealii/indent'
```
bobmyhill: dealii/indent?

Jenkinsfile
```diff
@@ -0,0 +1,145 @@
+#!groovy
+//
+// Jenkinsfile to control automated testing using https://jenkins.tjhei.info
+//
+// To run the tests locally, you can execute the following in a burnman working directory:
+//
+// docker run --rm -v "$(pwd):/src" tjhei/burnman:v4 /bin/bash -c "cd /src && ./test.sh"
```
bobmyhill: It might be worth adding that the user must first install docker and add themselves as a user.
`sudo apt install docker.io`
`sudo usermod -a -G docker $USER`

It's also definitely worth commenting that attempting to run the tester on a local machine will involve downloading 2 Gb of files...

Jenkinsfile
tjhei: yes, I need some docker image to run this in and this one works. :-)

We might need a different image if we actually want to format code here.

PR #350, opened 2019-02-03
added formula to composite

bobmyhill:
Added molar formula to composite. Also added formulae to endmembers in SLB_2005, and added a test for the new functionality.

### DIFF
--- a/burnman/composite.py
+++ b/burnman/composite.py
@@ -12,7 +12,7 @@ from .mineral import Mineral
 from . import averaging_schemes
 from . import chemicalpotentials
 from .solidsolution import SolidSolution
-
+from .processchemistry import sum_formulae
 
 def check_pairs(phases, fractions):
         if len(fractions) < 1:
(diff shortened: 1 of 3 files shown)


burnman/composite.py
```diff
@@ -199,6 +199,13 @@ def to_string(self):
         """
         return "{0}: {1}".format(self.__class__.__name__, self.name)
 
+    @material_property
+    def formula(self):
+        """
+        Returns molar chemical formula of the composite
+        """
+        return sum_formulae([ph.formula for ph in self.phases], self.molar_fractions)
```
tjhei: what is ``formula`` doesn't exist? Maybe we can produce a better error message?

bobmyhill: There are already error messages written for solidsolution and mineral, which pop up either during init (solidsolution) or when formula is called (mineral). Is there anywhere else we want errors?

tjhei: ah, yes. You are right. I didn't see that ``.formula`` here is a function call (and it will give an error for minerals without a formula).

PR #354, opened 2019-02-12
Clean main

sannecottaar:
I've been cleaning out main.py. Some things belonged in tools. This does mean that velocities_from_rock disappears, but we did deprecate that function almost four years ago. 

This branch still needs to run through the tests, so I might have to fix further things (the tests run into errors where the output looks different on my computer).


sannecottaar:
\rebuild

PR #357, opened 2019-04-23
Create build environment Dockerfiles

tjesser-ucdavis-edu:
I used `test.sh` to make sure the dependencies were correct. The installation guide didn't mention the python sympy library, but `test.sh` complained about not being able to import it. Is that just a `test.sh` dependency?

PR #361, opened 2019-04-27
removed handle/label calls to matplotlib

bobmyhill:
Addresses version problems raised in #357


bobmyhill:
@tjesser-ucdavis-edu This PR removes the offending matplotlib calls. continuous-integration/jenkins/pr-merge is still complaining about latex not existing, but hopefully jenkins should be happy now.

PR #362, opened 2019-04-29
Add texlive to bionic buildenvs

tjesser-ucdavis-edu:


bobmyhill:
@tjesser-ucdavis-edu jenkins still isn't happy, bless 'im.

tjesser-ucdavis-edu:
Jenkins uses the Docker containers on DockerHub and DockerHub is only set up to build the Dockerfiles in master. Since this pull request is for changing the buildenv container, the Jenkins builder won't reflect the change.

It's not a great dependency chain, but the build environment shouldn't change as much as other parts of the project.

Only other solution I can think of would be to make the package installation part of the build process, but that would mean downloading and installing packages for every PR.

bobmyhill:
Ah, I see. Thanks for the explanation!

PR #365, opened 2019-05-22
fixed scipy's factorial import issue

jrenaud90:
scipy version 1.3.0+ moved factorial from scipy.misc to scipy.special. This change will attempt to pull from misc (for older versions) and fall back to special.


bobmyhill:
@jrenaud90 This is great, thank you!

PR #367, opened 2019-07-02
avoid deprecation warnings

bobmyhill:
numpy and scipy continue to improve, and this is leading to a few function moves / deprecation warnings. The following line changes keep burnman up to date.


tjhei:
:+1:

PR #370, opened 2020-05-11
Fix to solve EoS with Gibbs energy constraint

S-Hunt:
Code needed to refer to the model not to 'self'


bobmyhill:
Hi @S-Hunt! Thanks for finding this! Could you remove the .DS_Store file from the PR? You could either amend and forc-epush the commit or do this from github.com by navigating to `Files changed` and deleting the offending file.

S-Hunt:
Hi @bobmyhill I should have deleted the .DS_Store now.

PR #373, opened 2021-03-15
Update Readme.md

bobmyhill:
Removed support for Python versions which have reached end of life (3.5 and earlier).

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -31,7 +31,11 @@ Authors (as of 2015, listed alphabetically by first name):
 
 * Python 3.6+
 * Python modules:
-  NumPy, SciPy, SymPy, Matplotlib, pycddlib
+  NumPy, SciPy, SymPy, Matplotlib
+  
+## Optional, needed for some functionality
+
+* pycddlib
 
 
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -29,9 +29,9 @@ Authors (as of 2015, listed alphabetically by first name):
 
 ## Requirements
 
-* Python 2.7.x or Python 3.4+
+* Python 3.6+
 * Python modules:
-  NumPy, SciPy, Matplotlib
+  NumPy, SciPy, SymPy, Matplotlib, pycddlib
```
tjhei, line 34: would it make sense to split between required and optional or do you think we should always require pycddlib?

bobmyhill: We can make pycddlib optional if you want. On a related note, your tester doesn't yet have pycddlib installed...

bobmyhill:
Ok, I've added an Optional category, and added pycddlib to it.

Readme.md
tjhei: We can just remove my jenkins instance, I would think. Objections?

PR #374, opened 2021-03-15
moved path hack to own file in examples directory

bobmyhill:
This PR removes the ugly hack adding the local burnman path to each of the example files, and instead puts it in its own file. This makes the preamble of all the files cleaner, but doesn't change the functionality in any way.

### DIFF
--- a/examples/burnman_path.py
+++ b/examples/burnman_path.py
@@ -1,0 +1,6 @@
+import os
+import sys
+
+# hack to allow scripts to be placed in subdirectories next to burnman:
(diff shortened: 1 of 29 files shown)


tjhei:
> Also removed some untidy whitespace.

Is this automated and if yes, how? We should enforce this using the tester.

bobmyhill:
I can probably set up an indent script like we do in ASPECT. I'll have a go tomorrow.

tjhei (approved):

PR #375, opened 2021-03-15
updated files to work with python 3.8.5

bobmyhill:
This PR updates files to work with python 3.8.5. There are several changes here:
- Warning messages have changed their syntax, so I've explicitly caught the warnings and printed their messages, rather than allowing python to print the messages itself. This should ensure version and system independent behaviour
- A couple of try/excepts were causing problems for the new warning messages. I've modified these so that they aren't bare excepts any more.
- numba/jit fallback is deprecated. We actually fell back to object mode for most of the asymmetric model jitted functions, because they use numpy functionalities which are not "jittable". As these function were vectorised anyway, I've just removed the jit decorations. This should not cause any significant slowdown for large jobs, and for small jobs will actually speed up the code.
- The imshow origin kwarg has changed its valid input for "low" to "lower".
- time.clock is deprecated, and is replaced with time.process_time.
- globally registered cmaps are now not editable and must be copied before modification.

### DIFF
--- a/burnman/solutionmodel.py
+++ b/burnman/solutionmodel.py
@@ -15,13 +15,12 @@ try:
     import os
     if 'NUMBA_DISABLE_JIT' in os.environ and int(os.environ['NUMBA_DISABLE_JIT']) == 1:
         raise ImportError("NOOOO!")
-    from numba import jit
+    from numba import njit
 except ImportError:
-    def jit(fn):
+    def njit(fn):
         return fn
 
 
-@jit
 def _ideal_activities_fct(molar_fractions, endmember_occupancies, n_endmembers, n_occupancies, site_multiplicities, endmember_configurational_entropies):
     site_occupancies = np.dot(molar_fractions, endmember_occupancies)
     a = np.power(site_occupancies,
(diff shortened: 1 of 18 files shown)


burnman/solutionmodel.py
```diff
@@ -13,32 +13,34 @@
...
     site_occupancies = np.dot(molar_fractions, endmember_occupancies)
     a = np.power(site_occupancies,
                  endmember_occupancies * site_multiplicities).prod(-1)
-    normalisation_constants = np.exp(endmember_configurational_entropies /
-                                     constants.gas_constant)
+    normalisation_constants = np.exp(endmember_configurational_entropies
+                                     / constants.gas_constant)
     return normalisation_constants * a
 
-@jit
```
tjhei, line 33: why did you remove this?

tjhei:
This is a large diff with many unrelated changes. Would you mind splitting things up? Otherwise I have a hard time reviewing things.

burnman/solutionmodel.py
bobmyhill: Because a) jit is deprecated (http://numba.pydata.org/numba-doc/latest/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit), b) these functions are already vectorised, so njit doesn't help much, and c) the rewrite is really fiddly, because numba doesn't yet allow operations over specific axes.

bobmyhill:
Sure, I can see what I can do.

bobmyhill:
Ok, this should be easier to review now. All the cosmetic changes have been shifted to #378

PR #382, opened 2021-03-19
Solution and bulk composition fitting

bobmyhill:
This PR adds new functions to fit analytical data:
- Calculating endmember fractions in a solution given a bulk composition
- Calculating phase fractions in an assemblage given phase and bulk compositions 

Both functions correctly apply non-negativity and equality constraints. The first function also allows the user to provide the variance-covariance matrix for the bulk composition.

- [x] Complete docstrings and add text to examples


bobmyhill:
Hi @tjhei, how do we add dependencies to the tester? 
cvxpy is the best way I've found to do constrained least squares in python: https://github.com/cvxpy/cvxpy. It's pretty powerful, and has several nice solvers.

 I'm happy for this to be an optional dependency like pycddlib.

tjhei:
> cvxpy is the best way I've found to do constrained least squares in python:

I fixed this in #396

bobmyhill:
The tester is still running 3.6.9...

gassmoeller:
It seems the new image was never build on dockerhub, and it seems dockerhub no longer supports automatic builds in general on free plans (I am sure it did that a while ago, but I cant find their announcement about the change). For now I build the new docker image manually and pushed it to docker hub. I triggered a rerun of the tester, and it has made it past the state where it crashed before. Long-term we may want to switch to github actions anyway, as we will be shutting down the Jenkins server at some point (I will look into this some time over the next year), but for now this should have fixed the tester.

bobmyhill:
Thanks @gassmoeller, thanks @tjhei!

PR #389, opened 2021-08-01
update docs to python 3

bobmyhill:
The old sphinx conf.py file does not work with python 3. 
This PR 
- updates that file,
- updates the master index file, and also
- removes some whitespace and duplicate keywords variables from ref.bib.


sphinx/conf.py
```diff
@@ -277,25 +192,32 @@
...
+    'preamble': preamble,
+    'maketitle': latex_maketitle
 }
 
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title, author, documentclass [howto/manual]).
 latex_documents = [
     ('index_pdf', 'BurnMan.tex', u'BurnMan Documentation',
-                  u'Sanne Cottaar, Timo Heister, Robert Myhill, Ian Rose, Cayman Unterborn', 'manual'),
+     u'Sanne Cottaar, Timo Heister, Robert Myhill, Ian Rose, Cayman Unterborn', 'manual'),
```
tjhei, line 217: Can you use the authors variable instead?

tjhei (approved):

bobmyhill:
This still isn't working; a lot of the docs are missing. I'll see if I can fix it tomorrow.

bobmyhill:
Ok, the autogenerated part of the manual is complete again. I needed to explicitly set the autodoc settings with some or all of the below (depending on case):
```
  :members:
  :undoc-members:
  :show-inheritance:
  :inherited-members:
```

tjhei:
I am not sure, but can you achieve the same by setting https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html instead of adding it to every class?

tjhei (approved):
Looks good!

PR #390, opened 2021-08-02
updating authorship orders

sannecottaar:
To reflect the current state of Bob being the main developer.


Readme.md
```diff
@@ -20,8 +20,8 @@ Source code: https://github.com/geodynamics/burnman
 
 Forums: https://community.geodynamics.org/c/burnman
 
-Authors (as of 2015, listed alphabetically by first name):
-* Bob Myhill
+Authors (as of 2021):
+* Bob (Robert) Myhill (main contributer)
```
bobmyhill, line 24:
```suggestion
* Bob (Robert) Myhill
```
I'm actually ok without the extra flag, but thank you for thinking of this :)

If we're removing the "listed alphabetically by first name" convention, should we adopt the same order as the author list in conf.py?

bobmyhill (commented):
This is absolutely not necessary, but thank you :)
I'm happy with this if everyone else is.

Unfortunately, we just merged #389 which creates a conflict in conf.py.

Readme.md
sannecottaar: I thought some variation is fine, as long as you are first. I think the extra flag is needed. Almost went ahead and added your email. It would be good if it was clear to people whom to approach if they did have suggestions or wanted to get involved.

bobmyhill: Ok, sure, I'm ok with that. In which case, there's a typo to fix ;)

bobmyhill, line 24:
```suggestion
* Bob (Robert) Myhill (main contributor)
```

bobmyhill (approved):

PR #400, opened 2021-08-04
added stoichiometric matrix and associated properties to Composite

bobmyhill:
This PR adds Composite properties which correspond to the stoichiometric matrix and associated column, row and nullspaces. These are useful for equilibrium calculations. 

This PR requires use of the pivots from Gaussian elimination. The necessary functions from sympy have been changing very quickly recently. For this PR, those functions are copied into reductions.py. We should be able to remove this code when sympy settles down. 

Secondly, the calculations are relatively expensive and rely on some optional parameters being set. For this reason, we only want to calculate the matrices once when they are first required, and then store the properties forever (i.e. we don't want to have to recalculate them every time we set_state or set_fractions). For this reason, I've also added a cached_property decorator. This can be replaced by the functools.cached_property decorator when we require Python 3.8+.

### DIFF
--- a/burnman/composite.py
+++ b/burnman/composite.py
@@ -12,26 +12,26 @@ import warnings
 from .material import Material, material_property, cached_property
 from .mineral import Mineral
 from . import averaging_schemes
-from . import chemicalpotentials
 from .solidsolution import SolidSolution
 from .processchemistry import sum_formulae, sort_element_list_to_IUPAC_order
 
+
 def check_pairs(phases, fractions):
-        if len(fractions) < 1:
-            raise Exception('ERROR: we need at least one phase')
+    if len(fractions) < 1:
+        raise Exception('ERROR: we need at least one phase')
 
-        if len(phases) != len(fractions):
-            raise Exception(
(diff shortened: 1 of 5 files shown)


burnman/composite.py
```diff
@@ -1,35 +1,37 @@
...
-        if len(fractions) < 1:
-            raise Exception('ERROR: we need at least one phase')
-
-        if len(phases) != len(fractions):
-            raise Exception(
-                'ERROR: different array lengths for phases and fractions')
 
-        total = sum(fractions)
-        if abs(total - 1.0) > 1e-10:
+def check_pairs(phases, fractions):
```
bobmyhill, line 19: this change is just fixing indentation

sannecottaar:
Looks good to me. I assume you have examples to commit to illustrate the use? The tests are slightly illustrative :)

bobmyhill:
I do have some examples in mind. Do you want them in this PR or the next?

sannecottaar:
Next is obviously fine.

PR #402, opened 2021-08-05
New example for Composite class

bobmyhill:
This PR adds an example for the Composite class. The example shows how to:
- initialize a composite
- interrogate it for different properties
- use the new properties introduced in #400 to show how to solve simple equilibrium problems (manually defining the unknowns - the general equilibrium class will come later).

It seems we already had a stray output file for an example_composite.py file that was never merged. This PR therefore overwrites that file with something that is actually used.

### DIFF
--- a/examples/example_composite.py
+++ b/examples/example_composite.py
@@ -150,13 +150,13 @@ if __name__ == "__main__":
     rock = burnman.Composite([andalusite, sillimanite, kyanite],
                              name='and-sill-ky assemblage')
 
-    def delta_affinity(x):
+    def delta_affinity_three_polymorphs(x):
         P, T = x
         rock.set_state(P, T)
         return rock.reaction_affinities
 
     print('Equilibrating the rock with fsolve')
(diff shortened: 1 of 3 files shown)


examples/example_composite.py
```diff
@@ -0,0 +1,239 @@
...
+        ol.set_composition([1. - x_fe_ol, x_fe_ol])
+        wad.set_composition([1. - x_fe_wad, x_fe_wad])
+        return rock.reaction_affinities
+
+    # First, get the univariant pressure at the Mg-rich end of the binary
+    # at a temperature of 1600 K.
+    T = 1600.
+    fo_mwd = burnman.Composite([burnman.minerals.SLB_2011.forsterite(),
+                                burnman.minerals.SLB_2011.mg_wadsleyite()])
+    brentq(delta_affinity_two_polymorphs, 12.e9, 16.e9, args=(T, fo_mwd))
```
sannecottaar, line 212: Should this be calling '_two_polymorphs', or the 'delta_affinity' just above?

PR #413, opened 2021-09-03
fix modified properties with pvt eos

bobmyhill:
In cases where the SLB dataset has a modified volume (e.g. quartz, which has a Landau transition with non-zero volume of disorder), Mineral needs to use the unmodified volume to calculate all of the unmodified properties, and then amend those with the modified properties.

This PR ensures that we do this correctly, and adds two more tests.

### DIFF
--- a/burnman/mineral.py
+++ b/burnman/mineral.py
@@ -139,7 +139,7 @@ class Mineral(Material):
     @material_property
     @copy_documentation(Material.molar_gibbs)
     def molar_gibbs(self):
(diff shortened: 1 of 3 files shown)


sannecottaar:
Can this not be hidden in self.molar_volume... checking if an unmodified version needs to be used or not. Only trying to avoid the additional tag, although it might just be my ignorance that makes it more confusing.

bobmyhill:
Unfortunately not. The method.property functions are intended to return the raw equation of state properties, so we should never use the "modified" volume as input to any self.method.property(P, T, V), as that is inconsistent with the EoS.

PR #414, opened 2021-09-03
add github actions

tjhei:


tjhei:
this seems to be working, @bobmyhill .
One of the tests fails though:
https://github.com/geodynamics/burnman/pull/414/checks?check_run_id=3513587761#step:5:20

bobmyhill:
Thanks @tjhei! This looks great. The broken test was due to sympy changing the format of its nullspace object. A rebase should fix the problem.

bobmyhill:
I've just run the tests with exactly the same package versions, so everything should be updated now. One more rebase should do it!

tjhei:
let's see!

bobmyhill:
Closer! I'm not going to make any wild claims this time, but the next set of deprecated calls have been removed.

tjhei:
Are you pushing directly to master? You could also just add commits here...

bobmyhill:
I was, not best practise, I know. Thankfully the tester now passes. Thanks for doing this!

PR #416, opened 2021-09-04
add requirements.txt, use in CI

tjhei:
- add basic requirements.txt
- update readme
- use requirements.txt in CI

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,8 @@ jobs:
          texlive \
          texlive-latex-extra
         python -m pip install --upgrade pip
-        pip install matplotlib numpy scipy sympy pycddlib cvxpy
+        python -m pip install -r requirements.txt
+        pip install pycddlib # this is optional
         python --version
     - name: test
       run: |
--- a/Readme.md
+++ b/Readme.md
@@ -31,7 +31,7 @@ Authors (as of 2021):
 
 * Python 3.7+
 * Python modules:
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -23,7 +23,8 @@ jobs:
          texlive \
          texlive-latex-extra
         python -m pip install --upgrade pip
-        pip install matplotlib numpy scipy sympy pycddlib cvxpy
+        python -m pip install -r requirements.txt
```
bobmyhill, line 26: we already have sphinx/pip_requirements.txt.

I'd suggest that we remove that file, rename the new requirements.txt to pip_requirements.txt, and change the line in test.sh from
`$PYTHON -m pip freeze | grep -f sphinx/pip_requirements.txt`
to
`$PYTHON -m pip freeze | grep -f ./pip_requirements.txt`

tjhei:
You know that ./requirements.txt is a common if not standard way to do things, right? The sphinx requirements are different, do you really want to merge them?

bobmyhill:
Yes, I know that a requirements.txt file is a standard way to check requirements (this is why I added the sphinx/pip_requirements.txt file back in 2017). The existing file is only used by ReadTheDocs. 

I'm not sure it would matter if we had all the requirements (code + docs) in a single file, but if you'd like to keep two separate files, I'm not going to insist on a merger. I just thought it would be nice to minimise file duplication.

tjhei:
I don't feel strongly about it, so feel free to change things!

PR #422, opened 2021-10-03
BurnMan 1.0 restructuring and documentation

bobmyhill:
This PR:
- Moves the Material classes into their own subdirectory
- Modifies the imports and doc strings accordingly
- Moves the deprecated misc/mineral_helpers.py into CHRU2014 (it should no longer be readily available to users)
- Fixes a couple of broken imports
- Removes some starred imports
- Adds burnman_path files to misc and misc/benchmarks, so that we can remove the nasty import hack
- Adds Composition class and functions to the docs
- Adds a couple of missing examples (fit_compositions, anisotropic_mineral) to the docs


sannecottaar:
Are layer.py and planet.py also going in classes?

bobmyhill:
Not in this PR (which is already too bulky!), but yes :)

PR #428, opened 2021-10-05
clean out misc

sannecottaar:
Cleaned out files in misc/ref that weren't used.
Removed pyrolite_uncertainty.py in misc folder
moved colors.py and helpers_solid_solution.py to contrib/CHRU2014 (only place it is used)
added readme to contrib/CHRU2014


bobmyhill:
@sannecottaar, I deal with the deprecated model.py stuff in #427, so you don't need to worry about removing that (the cause of your failing tests).

contrib/CHRU2014/paper_averaging.py
```diff
@@ -42,7 +42,7 @@
 
 import burnman
 from burnman import minerals
-import misc.colors as colors
+import .colors as colors
```
bobmyhill, line 45: I think `from contrib.CHRU2014 import colors` would work (here and in other scripts).

bobmyhill:
Ooops, `colors.py` is deleted from `misc/`, but not yet added into `contrib/CHRU2014/`. After that, everything *should* work :)

misc/ref/example_solution_creation_and_manipulation.py.out
```diff
@@ -1,156 +0,0 @@
-Endmember occupancies for bridgmanite in the FMASO system
```
bobmyhill, line 1: I think this file needs to be restored.

sannecottaar: I don't see the example?

bobmyhill: It's in contrib/solution_polytope. It's not a perfect name, but it's the name given in the published paper.

PR #445, opened 2021-10-29
update 1D ASPECT profile notebook

bobmyhill:
This PR overhauls the old 1D ASPECT profile notebook so that it works again after the major changes to jupyter/ipython over the last few years.

For some reason, the interactive widgets don't work on binder ("Error displaying widget: model not found" seems to be a common problem with no single solution), so I've also made it easy to run the script to produce a static figure instead (by not running the first code block).


ianhi:
Hi @bobmyhill  you can consistenly get the interactive widgets working if you add a binder config that will install ipympl. An example commit that does this is:

https://github.com/ianhi/mpl-image-labeller/commit/1b0b5d05256846b683e08309a3fb270803ea9849

ianhi:
The postbuild would also let you simplify this installation step to just be import:

image

PR #500, opened 2022-09-13
Update paper.md

jdannberg:
Some more things I noticed:

Figure 2: I noticed that these properties are not the one mentioned in the text as being important. Would it make sense to add a sentence in the figure caption what the excess Gibbs energy and the endmember activities are useful for?

Figure 3: Would it be worth mentioning the source of this data? I also don’t see the orange data points. 

Figure 4: I am sure the there’s detailed documentation on the planet builder somewhere, would it make sense to link to that here? From the caption alone, it does not really become clear what the planet builder does exactly.

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -77,16 +77,11 @@ are not accessible using existing codes.
 When `BurnMan` was first released [@Cottaar:2014], its focus was on
 the seismic properties of the lower mantle, using a single endmember mineral
 database [@Stixrude:2011] as a foundation. Since then,
-its scope has expanded considerably. Some workflows of the current code
-are provided in \autoref{fig:workflows}. 
-
-![Some possible BurnMan workflows. Input/output shown as ellipses, functions as
-rectangles and classes as diamonds.\label{fig:workflows}](figures/workflows.png)
-
+its scope has expanded considerably. 
 `BurnMan` now contains equations of state
 for minerals and melts from several published datasets. A model for anisotropy
 has recently been added. Several solution model formulations are included
-(\autoref{fig:garnetsolution}),
+(with one example model shown in \autoref{fig:garnetsolution}),
 and a composite material model that allows several seismic averaging schemes.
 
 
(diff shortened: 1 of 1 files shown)


paper/paper.md
```diff
@@ -77,16 +77,11 @@ are not accessible using existing codes.
 When `BurnMan` was first released [@Cottaar:2014], its focus was on
 the seismic properties of the lower mantle, using a single endmember mineral
 database [@Stixrude:2011] as a foundation. Since then,
-its scope has expanded considerably. Some workflows of the current code
-are provided in \autoref{fig:workflows}. 
-
-![Some possible BurnMan workflows. Input/output shown as ellipses, functions as
-rectangles and classes as diamonds.\label{fig:workflows}](figures/workflows.png)
-
+its scope has expanded considerably.
```
jdannberg, line 80: This is just a suggestion, but I wondered if it makes sense to remove Figure 1. In the original paper draft on overleaf, you’re using it to explain the individual classes, but without that, I find it a bit confusing, and the workflows that are worth mentioning are already listed in the text.

paper/paper.md
```diff
@@ -77,16 +77,11 @@ are not accessible using existing codes.
...
-
-![Some possible BurnMan workflows. Input/output shown as ellipses, functions as
-rectangles and classes as diamonds.\label{fig:workflows}](figures/workflows.png)
-
+its scope has expanded considerably. 
 `BurnMan` now contains equations of state
 for minerals and melts from several published datasets. A model for anisotropy
 has recently been added. Several solution model formulations are included
-(\autoref{fig:garnetsolution}),
+(with one example model shown in \autoref{fig:garnetsolution}),
```
jdannberg, line 84: Because the text says “Several solution model formulations are included (Figure 2)” but then Figure 2 only shows “a published model”.

bobmyhill:
@jdannberg Thanks! I'll make the other changes you suggested now.

Issue #524, opened 2023-05-11
Burnman readable files


vivianotieno:
Dear Burnman Community

I have been trying to use the newer version of burnman to create a burnman readable file, and I keep getting the below error: I am using the more recent version of perple_x of v.7.0.9. I have tried the various versions of burnman, and non seem to work. The error comes from the perplex.py file. I have run the Werami file, and the issue might be due to an additional prompt not in the previous versions. Find below are the resulting prompts from running the create-burnman table: 

Keeping P-T range the same as the original project range.

Perple_X release 7.0.9, March 27, 2023.

Copyright (C) 1986-2023 James A D Connolly <www.perplex.ethz.ch/copyright.html>.

Enter the project name (the name assigned in BUILD) [default = my_project]:

Reading Perple_X options from: stx21_perplex_option.dat
Writing Perple_X option summary to: not requested

Perple_X computational option settings for WERAMI:

Keyword:               Value:     Permitted values [default]:

Input/Output options:

aqueous_output          T         [F] T
aqeuous_species         20        [20] 0-150
aq_solvent_composition  y         [y] m: y => mol fraction, m => molality
aq_solute_composition   m         y [m]: y => mol fraction, m => molality
spreadsheet             T         [F] T
logarithmic_p           F         [F] T
logarithmic_X           F         [F] T
bad_number             NaN        [NaN]
composition_constant    F         [F] T
composition_phase       mol       [mol] wt
composition_system      wt        [wt] mol
proportions             vol       [vol] wt mol
absolute                F         [F] T
cumulative              F         [F] T
fancy_cumulative_modes  F         [F] T
interpolation           on        [on] off 
melt_is_fluid           T         [T] F
solution_names          mod       [model] abbreviation full
structural_formulae     T         [T] F
output_species          T         [T] F
output_species_props    F         [F] T
seismic_output          som       [some] none all
pause_on_error          T         [T] F
poisson_test            F         [F] T
interim_results         aut       [auto] off manual
sample_on_grid          T         [T] F

Information file output options:

option_list_files       F         [F] T; echo computational options

Thermodynamic options:

approx_alpha            T         [T] F
Anderson-Gruneisen      F         [F] T
hybrid_EoS_H2O             4      [4] 0-2, 4-7
hybrid_EoS_CO2             4      [4] 0-4, 7
hybrid_EoS_CH4             0      [0] 0-1, 7
fd_expansion_factor     2.0       [2] >0
[...]

PR #537, opened 2023-06-16
added useful PerpleX messages

bobmyhill:
PerpleX 7.x has slightly changed the functionality and expected user input of werami. This PR alerts the user to the following things they must now do to prepare their directory before running the burnman.perplex.create_perplex_table():
- run vertex
- set sample_on_grid to F in the perplex option file if they want to use the default P-T range

The code itself is reformatted slightly to allow easier processing of the stdout from werami.

This PR is a first step toward addressing #524. 

The code works with Perple_X release 7.0.9.


vivianotieno:
Dear Bobmyhill

I'm sorry for not getting back to you sooner. The new version works. I didn't have to change the sample grid mode.

PR #600, opened 2024-10-06
enable pycddlib3

bobmyhill:
The update of pycddlib to Version 3 comes with several improvements:
- a standalone package (pycddlib-standalone) for float representations of polytopes that doesn't require non-python dependencies that can be difficult to install
- a separation of float and fraction modules that makes it easier to type check.
- several new convenience functions.

One downside of the new version is that the old github workflow for installation of the full pycddlib is broken, and I can't figure out how to make the necessary changes.

In this PR, I've added pycddlib-standalone as a dependency, and changed the polytope functions to allow either pycddlib-standalone or the full pycddlib module. If cdd.gmp (the Fractions module) is not found, any calculations that would have used that module instead use cdd (the floats module).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on: [push, pull_request, workflow_dispatch]
 jobs:  
   indent:
     name: indent
-    runs-on: [ubuntu-20.04]
+    runs-on: [ubuntu-24.04]
 
     strategy:
       matrix:
@@ -34,7 +34,7 @@ jobs:
 
   linux:
     name: test
-    runs-on: [ubuntu-20.04]
+    runs-on: [ubuntu-24.04]
 
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


mattiejastje:
Swell! I was a bit nervous to release this since it's a big API change. But hopefully for the better in the long term.

I don't know if you managed this yet, but to get things working for pycddlib (the non-standalone version) on the github actions workflow, on Ubuntu it should be enough to simply install gmp and cddlib itself from apt:

```
sudo apt-get install -y libcdd-dev libgmp-dev
```

Pycddlib should then find the headers and library and use those.

bobmyhill:
It was a big API change, but your changelog was really helpful. The changes also allowed me to streamline some calculations and isolate some odd type behaviour.

I did try your apt-get suggestion this morning, and it failed, but trying again this afternoon it works. My guess is the update to using ubuntu 24.04 in the tester (rather than 20.04) fixed the issue.

Issue #620, opened 2025-04-24
Non-site elements not added to solution_formulae in solution models


bjheinen:
The function process_solution_chemistry from utils/chemistry.py sets up site occupancy and chemical formulae for a given solution model. The snippet from line 403 (`# Loop over species after site`) looks like it should add any chemical species not on a solution site to the chemical formula dictionary (solution_formulae), which is an attribute of the given SolutionModel. A slight bug means this isn't working at the moment:

```
>>> sol_a = bm.classes.solutionmodel.IdealSolution(endmembers = [[bm.classes.mineral.Mineral(),'[Mg]3[Al]2'],[bm.classes.mineral.Mineral(), '[Mg]3[Mg1/2Si1/2]2']])
>>> sol_b = bm.classes.solutionmodel.IdealSolution(endmembers = [[bm.classes.mineral.Mineral(),'[Mg]3[Al]2Mg3Al3Si3'],[bm.classes.mineral.Mineral(), '[Mg]3[Mg1/2Si1/2]2Mg3Al3Si3']])

>>> sol_a.solution_formulae
[{'Mg': 3.0, 'Al': 2.0}, {'Mg': 4.0, 'Si': 1.0}]
>>> sol_b.solution_formulae
[{'Mg': 3.0, 'Al': 2.0}, {'Mg': 4.0, 'Si': 1.0}]
```

Doesn't look like solution_formulae is used anywhere else, so this snippet could be removed instead of fixed.

bobmyhill:
Thanks @bjheinen , I merged #621, but I think for the sake of code clarity it is better to remove the object and sanitise existing naming (#622).

bobmyhill:
Thanks for your help!

bjheinen:
No problem! @bobmyhill That sounds sensible

PR #621, opened 2025-04-24
Ensure not_in_site is string and not filter

bjheinen:
This PR fixes issue #620. 
re.findall fails silently on a filter (returned by `str(filter(None, site_split[1]))`)
solution_formulae doesn't seem to be used anywhere else in the codebase, so section could be removed instead of merging this PR.


bobmyhill:
Thanks @bjheinen , I'll merge this now, since you should be given credit for finding and fixing a bug!
In another PR I'll remove this bit of the function as the output object is not guaranteed to make any sense.

PR #690, opened 2025-10-16
move regularization

bobmyhill:
This pull request simplifies the regularization logic in the Newton step.

**Regularization improvements:**

* Changed the default value for the `regularization` parameter from `0.0` to `np.finfo(float).eps`.
* Updated the docstring to reflect the new default for `regularization`, specifying it now defaults to numpy float epsilon.

**Solver logic updates:**

* Moved the regularization of ill-conditioned Jacobian matrices earlier in the solver logic to be more consistent throughout the calculations.
* The final adjustment step now recomputes the solution vector and Jacobian without regularization to ensure tolerances are satisfied.

### DIFF
--- a/burnman/optimize/nonlinear_solvers.py
+++ b/burnman/optimize/nonlinear_solvers.py
@@ -722,11 +722,6 @@ class DampedNewtonSolver:
             dx = lu_solve(luJ, -sol.F)
             dx_norm = np.linalg.norm(dx, ord=2)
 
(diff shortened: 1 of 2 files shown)


bobmyhill:
@bjheinen seems fine. There's still at least one test that is intended to fail but can apparently fail with two different errors. I'll merge this for now and deal with the failing case whenever the testers next decide to reveal it to us.

